PR #2, opened 2015-02-16
Fixed problem with response.error not being of NSError class

antisoftware:
Fixed issue with the "error" value (KVO) being overwritten by a generic assignment in case of textual error returned by Syncano: the "checkAndSetError" method was correctly assigning NSError object while the generic assignment made afterwards was wiping this out and replacing with NSString


Syncano/Responses/SyncanoResponse.m
```diff
@@ -64,7 +64,8 @@ + (instancetype)responseFromJSON:(NSDictionary *)json {
 		jsonContent = json[@"data"];
 	}
 	for (NSString *key in [jsonContent allKeys]) {
-		[response setValue:jsonContent[key] forKey:key];
+    if (![key isEqualToString:@"error"])
```
atteroTheGreatest: What an ugly indentation ;). It wouldn't work in python.

atteroTheGreatest:
Amazing. LGTM ;)))

PR #3, opened 2015-03-17
another portion of changes

antisoftware:
changed APIClient to call [self class] instead of [APIClient class]
changed nullSafe to SyncanoNullSafe category name for NSObject
corrected #include to #import "SyncanoArray.h"
added proper keys to Syncano.m to avoid pure strings
added check for the existence of Syncano.plist in custom project
used Mantle to deserialize JSON to objects
introduced SyncanoError and SyncanoException
changed protocol name from SyncanoObjectProtocol to SyncanoObject
changed getter methods of SyncanoObject protocol into properties

### DIFF
--- a/Syncano/APIClient.m
+++ b/Syncano/APIClient.m
@@ -10,7 +10,7 @@
 
 @implementation APIClient
 
-static NSString *const kSyncanoBaseURL = @"https://syncanotest1-env.elasticbeanstalk.com:443/v1/";
+static NSString *const kSyncanoBaseURL = @"https://syncanotest1-env.elasticbeanstalk.com:443/";
 
 + (instancetype)sharedClient {
 	static APIClient *_sharedClient = nil;
@@ -20,7 +20,8 @@ static NSString *const kSyncanoBaseURL = @"https://syncanotest1-env.elasticbeans
 		_sharedClient.securityPolicy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModeNone];
     _sharedClient.securityPolicy.allowInvalidCertificates = YES; // TODO
 		_sharedClient.requestSerializer = [AFJSONRequestSerializer serializerWithWritingOptions:NSJSONWritingPrettyPrinted];
-		_sharedClient.responseSerializer = /*_sharedClient.jsonOutput ? [AFJSONResponseSerializer serializer] : */[AFHTTPResponseSerializer serializer];
+    _sharedClient->_jsonOutput = YES;
+		_sharedClient.responseSerializer = (AFHTTPResponseSerializer <AFURLResponseSerialization> *)(_sharedClient.jsonOutput ? [AFJSONResponseSerializer serializer] : [AFJSONRequestSerializer serializer]);
 	});
 	return _sharedClient;
 }
--- a/Syncano/APIClient.m
+++ b/Syncano/APIClient.m
@@ -15,7 +15,7 @@
 	static dispatch_once_t onceToken;
 	dispatch_once(&onceToken, ^{
 		NSDictionary *config = [NSDictionary dictionaryWithContentsOfURL:[NSURL fileURLWithPath:[[NSBundle mainBundle] pathForResource:@"Config" ofType:@"plist"]]];
-		_sharedClient = [[APIClient alloc] initWithBaseURL:[NSURL URLWithString:config[@"BaseURL"]]];
+		_sharedClient = [[self alloc] initWithBaseURL:[NSURL URLWithString:config[@"BaseURL"]]];
 		_sharedClient.securityPolicy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModeNone];
 		_sharedClient.requestSerializer = [AFJSONRequestSerializer serializerWithWritingOptions:NSJSONWritingPrettyPrinted];
 		_sharedClient.responseSerializer = _sharedClient.jsonOutput ? [AFJSONResponseSerializer serializer] : [AFHTTPResponseSerializer serializer];
--- a/Syncano/APIClient.m
+++ b/Syncano/APIClient.m
@@ -10,15 +10,17 @@
 
 @implementation APIClient
 
+static NSString *const kSyncanoBaseURL = @"https://syncanotest1-env.elasticbeanstalk.com:443/v1/";
+
 + (instancetype)sharedClient {
 	static APIClient *_sharedClient = nil;
 	static dispatch_once_t onceToken;
 	dispatch_once(&onceToken, ^{
-		NSDictionary *config = [NSDictionary dictionaryWithContentsOfURL:[NSURL fileURLWithPath:[[NSBundle mainBundle] pathForResource:@"Config" ofType:@"plist"]]];
-		_sharedClient = [[self alloc] initWithBaseURL:[NSURL URLWithString:config[@"BaseURL"]]];
+		_sharedClient = [[self alloc] initWithBaseURL:[NSURL URLWithString:kSyncanoBaseURL]];
 		_sharedClient.securityPolicy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModeNone];
+    _sharedClient.securityPolicy.allowInvalidCertificates = YES; // TODO
 		_sharedClient.requestSerializer = [AFJSONRequestSerializer serializerWithWritingOptions:NSJSONWritingPrettyPrinted];
-		_sharedClient.responseSerializer = _sharedClient.jsonOutput ? [AFJSONResponseSerializer serializer] : [AFHTTPResponseSerializer serializer];
+		_sharedClient.responseSerializer = /*_sharedClient.jsonOutput ? [AFJSONResponseSerializer serializer] : */[AFHTTPResponseSerializer serializer];
 	});
 	return _sharedClient;
 }
--- a/Syncano/NSObject+nullSafe.h
+++ b/Syncano/NSObject+nullSafe.h
@@ -12,6 +12,6 @@
 
 @interface NSObject (SyncanoNullSafe)
 
-- (id)nullSafe;
+- (id)syncanoNullSafe;
 
 @end
--- a/Syncano/NSObject+nullSafe.h
+++ b/Syncano/NSObject+nullSafe.h
@@ -1,6 +1,6 @@
 //
 //  NSObject+nullSafe.h
-//  LoaderSplitView
+//  Syncano
 //
 //  Created by Mateusz on 22.01.2015.
 //  Copyright (c) 2015 Syncano. All rights reserved.
@@ -10,7 +10,7 @@
 
 #define ASSIGN_IF_NOT_NIL(leftOperand, rightOperand) { if (rightOperand) leftOperand = rightOperand; }
 
-@interface NSObject (nullSafe)
+@interface NSObject (SyncanoNullSafe)
 
 - (id)nullSafe;
 
--- a/Syncano/Syncano.m
+++ b/Syncano/Syncano.m
@@ -58,7 +58,7 @@ static Syncano *_sharedInstance = nil;
       defaultName = [Syncano settings][kDefaultName];
     
     if ([defaultAPIKey length] && [defaultName length]) {
-      _sharedInstance = [[Syncano alloc] initWithAPIKey:defaultAPIKey name:defaultName];
+      _sharedInstance = [[self alloc] initWithAPIKey:defaultAPIKey name:defaultName];
     }
   });
   
@@ -99,7 +99,7 @@ static Syncano *_sharedInstance = nil;
 }
 
 + (instancetype)syncanoWithAPIKey:(NSString *)apiKey name:(NSString *)name {
-  return [[Syncano alloc] initWithAPIKey:apiKey name:name];
+  return [[self alloc] initWithAPIKey:apiKey name:name];
 }
 
 - (void)dealloc {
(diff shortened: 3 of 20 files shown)


Syncano/APIClient.m
```diff
@@ -15,7 +15,7 @@ + (instancetype)sharedClient {
 	static dispatch_once_t onceToken;
 	dispatch_once(&onceToken, ^{
 		NSDictionary *config = [NSDictionary dictionaryWithContentsOfURL:[NSURL fileURLWithPath:[[NSBundle mainBundle] pathForResource:@"Config" ofType:@"plist"]]];
-		_sharedClient = [[APIClient alloc] initWithBaseURL:[NSURL URLWithString:config[@"BaseURL"]]];
+		_sharedClient = [[self alloc] initWithBaseURL:[NSURL URLWithString:config[@"BaseURL"]]];
```
lifcio: :+1:

Syncano/NSObject+nullSafe.h
```diff
@@ -10,7 +10,7 @@
 
 #define ASSIGN_IF_NOT_NIL(leftOperand, rightOperand) { if (rightOperand) leftOperand = rightOperand; }
 
-@interface NSObject (nullSafe)
+@interface NSObject (SyncanoNullSafe)
 
 - (id)nullSafe;
```
lifcio: Sorry If I was not being clear here before - the issue I raised before was about adding the prefix to actually method implemented in the category of existing object. So e.g instead of
```objc
- (id)nullSafe;
```
it could be
```objc
- (id)syn_nullSafe;
```
Again, this is not a big issue, but it is (maybe not well known) [standard Apple recommends](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html) to avoid potential category methods name clashes.

Syncano/Syncano.m
```diff
@@ -25,103 +36,111 @@ @implementation Syncano
...
-	return _sharedInstance;
+  static dispatch_once_t onceToken;
+  
+  __block NSString *defaultAPIKey = apiKey;
+  __block NSString *defaultName = name;
+  dispatch_once(&onceToken, ^{
+    if (![defaultAPIKey length])
+      defaultAPIKey = [Syncano settings][kAPIKey];
+    if (![defaultName length])
+      defaultName = [Syncano settings][kDefaultName];
```
lifcio: :+1: for constants

Syncano/Syncano.m
```diff
@@ -25,103 +36,111 @@ @implementation Syncano
...
+    
+    if ([defaultAPIKey length] && [defaultName length]) {
```
lifcio: @atteroTheGreatest Do we have standardized API key length? Was it 40?
If we do, than probably it would be better to check for it here explicitly.

Syncano/Syncano.m
```diff
@@ -25,103 +36,111 @@ @implementation Syncano
...
+      _sharedInstance = [[Syncano alloc] initWithAPIKey:defaultAPIKey name:defaultName];
```
lifcio: Here is the same issue as before with APIClient, it's better to use `[self alloc]`. It's even more valid here, because Syncano class probably will be at some point subclassed by developers.

Syncano/Syncano.m
```diff
@@ -25,103 +36,111 @@ @implementation Syncano
...
-	if (!instance)
-		return [Syncano sharedInstanceWithAPIKey:[Syncano settings][@"apiKey"] name:name];
-
-	instance->_name = [name copy];
-	return instance;
+  Syncano *instance = [Syncano sharedInstance];
+  if (!instance)
+    return [Syncano sharedInstanceWithAPIKey:[Syncano settings][kAPIKey] name:name];
+  
+  instance->_name = [name copy];
```
lifcio: When you can, you should try to avoid using variables directly. It's much safer and less error prone to use properties when suitable.
E.g. both `name` and `apiKey` could be transformed into
```objc
@interface Syncano ()
@property (copy,readwrite) NSString *apiKey;
@property (copy,readwrite) NSString *name;
@end
```
and you could then change following lines:
```objc
instance->_name = [name copy];
_apiKey = [apiKey copy];
_name = [name copy];
```
into
```objc
instance.name = name;
self.apiKey = apiKey;
self.name = name;
```
Not to mention it's right now much easier to change it from `copy` to `retain` if it was even needed.

Syncano/Syncano.m
```diff
@@ -25,103 +36,111 @@ @implementation Syncano
...
-		_name = [name copy];
-	}
-	return self;
+  if (![apiKey length] || ![name length])
+    return nil;
+  
+  self = [super init];
+  if (self) {
+    _apiKey = [apiKey copy];
+    _name = [name copy];
```
lifcio: As commented above - please change into using properties.

Syncano/Syncano.m
```diff
@@ -25,103 +36,111 @@ @implementation Syncano
...
 }
 
 + (instancetype)syncanoWithName:(NSString *)name {
-	return [Syncano syncanoWithAPIKey:[Syncano settings][@"apiKey"] name:name];
+  return [Syncano syncanoWithAPIKey:[Syncano settings][kAPIKey] name:name];
 }
 
 + (instancetype)syncanoWithAPIKey:(NSString *)apiKey name:(NSString *)name {
-	return [[Syncano alloc] initWithAPIKey:apiKey name:name];
+  return [[Syncano alloc] initWithAPIKey:apiKey name:name];
```
lifcio: For the class methods go with `[self alloc]` instead of `[Syncano alloc]`.

Syncano/Syncano.m
```diff
@@ -25,103 +36,111 @@ @implementation Syncano
...
 }
 
 - (void)dealloc {
-	_apiKey = nil;
-	_name = nil;
+  _apiKey = nil;
+  _name = nil;
```
lifcio: When `apiKey` and `name` are change into properties you don't even need to nil them here - `dealloc` will be empty and beautiful ;)

Syncano/Syncano.m
```diff
@@ -25,103 +36,111 @@ @implementation Syncano
...
-																		else if (failure)
-																			failure(task, [NSError errorWithDomain:@"Syncano" code:100 userInfo:nil]);
-																	}
-																}
-																failure:^(NSURLSessionDataTask *task, NSError *error) {
-																	if (failure)
-																		failure(task, error);
-																}];
+  if (![class conformsToProtocol:@protocol(SyncanoObject)]) {
+    [[SyncanoException exceptionWithCode:SyncanoExceptionInvalidParameter reason:[NSString stringWithFormat:@"%@ does not conform to protocol %@", NSStringFromClass(class), NSStringFromProtocol(@protocol(SyncanoObject))]] raise];
```
lifcio: :+1: 
In the future, in one of the next iterations we should probably also consider having those strings in `.strings` file(s).

Syncano/SyncanoError.h
```diff
@@ -0,0 +1,21 @@
...
+//  Created by Mateusz on 17.03.2015.
+//  Copyright (c) 2015 Syncano. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+typedef enum {
+  SyncanoErrorInvalidParameter = 0,
+  SyncanoErrorResponseIsNotJSONDictionary
+} SyncanoErrorCode;
```
lifcio: You could switch to [modern Objective-C enumeration technique](https://developer.apple.com/library/ios/releasenotes/ObjectiveC/ModernizationObjC/AdoptingModernObjective-C/AdoptingModernObjective-C.html).
E.g. here, it would be 
```objc
typedef NS_ENUM(NSInteger, SyncanoErrorCode) {
    SyncanoErrorInvalidParameter = 0,
    SyncanoErrorResponseIsNotJSONDictionary
};
```

Syncano/SyncanoException.h
```diff
@@ -0,0 +1,22 @@
...
+//
+//  Created by Mateusz on 17.03.2015.
+//  Copyright (c) 2015 Syncano. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+extern NSString *const kSyncanoExceptionInvalidParameter;
+
+typedef enum {
```
lifcio: As above, could be instead:
```objc
typedef NS_ENUM(NSInteger, SyncanoExceptionCode) {
    SyncanoExceptionInvalidParameter = 0,
};
```

Syncano/SyncanoException.m
```diff
@@ -0,0 +1,42 @@
...
+  switch (code) {
+    case SyncanoExceptionInvalidParameter:
+      name = kSyncanoExceptionInvalidParameter;
+      break;
+      
+    default:
+      break;
+  }
+  
+  return [[[self class] alloc] initWithName:name reason:reason userInfo:userInfo];
```
lifcio: If it's a class method, you don't need `[[self class] alloc]`, you can do 
```
[[self alloc] init...]
```
instead. Practically same effect, but shorter notation.

Syncano/SyncanoException.m
```diff
@@ -0,0 +1,42 @@
...
+
++ (instancetype)exceptionWithCode:(SyncanoExceptionCode)code {
+  NSString *reason = nil;
+  
+  switch (code) {
+    default:
+      break;
+  }
+  
+  return [[self class] exceptionWithCode:code reason:reason];
```
lifcio: If this is calling the other method anyway, then it's probably better to just keep the switch in one place.
Is there a reason to have it in both? Would they do something else?

Syncano/SyncanoObject.m
```diff
@@ -27,4 +27,10 @@ - (id)initWithJSON:(NSDictionary *)json {
 	return self;
 }
 
+#pragma mark - MTLJSONSerializing
+
+- (NSString *)JSONKeyPathForPropertyKey:(NSString *)key {
+  return key; // by default Syncano fields and object properties are presumed to match
```
lifcio: This is not true for the building fields - the basic ones. E.g. on Syncano there is 'created_at` and `updated_at`.
And of course, on Syncano there is no `dbID`, so Syncano's `id` will need to be translated into it too.

Syncano/SyncanoObject.h
```diff
@@ -17,8 +18,8 @@
 	NSString *_sidekick;
 }
 
-@property (strong, readonly) NSDate *createdAt;
-@property (strong, readonly) NSDate *updatedAt;
+@property (copy, readonly) NSDate *createdAt;
+@property (copy, readonly) NSDate *updatedAt;
 @property (assign, readonly) NSInteger revision;
 @property (assign, readonly) NSInteger dbID;
 @property (copy, readonly) NSString *sidekick;
```
lifcio: What's `sidekick`? Is this something on Syncano side? I'm not familiar with this parameter.

lifcio:
@antisoftware Could you address the issues listed?

Syncano/Syncano.m
antisoftware: I wouldn't agree with that. What if you change it in the future? We should keep it simple and pass whatever user wants us to pass. If the APIKey is wrong the result will still be in accordance with that (empty).
The only reason this would make sense would be to minimize the traffic from mobile apps to the server. Still, the question I raised at the beginning is valid.

Syncano/SyncanoException.m
antisoftware: Theoretically you might want to do sth in the "exceptionWithCode:" method that depends solely on the code while of course the "exceptionWithCode:reason:" version also wants to behave conditionally. I understand this way it's not the fastest possible, but keeps the full potential of using both methods individually.

Syncano/NSObject+nullSafe.h
antisoftware: ok

Syncano/Syncano.m
antisoftware: agreed

Syncano/Syncano.m
antisoftware: matter of taste, but I will follow your suggestions

Syncano/Syncano.m
antisoftware: as above

Syncano/Syncano.m
antisoftware: sure

Syncano/Syncano.m
antisoftware: agree, see my previous comments

Syncano/SyncanoError.h
antisoftware: ok

Syncano/SyncanoException.h
antisoftware: ok

Syncano/SyncanoException.m
antisoftware: again, purely a notation matter, but fine

Syncano/SyncanoObject.m
antisoftware: sure, it was just a temporary solution

Syncano/SyncanoObject.h
antisoftware: I guess it was shown for a test object and I didn't notice it was added by someone else. To be removed of course.

Syncano/Syncano.m
atteroTheGreatest: If key is invalid, you will get a permission error from the api. No need to check it everywhere.

lifcio: Ok make sense, then yeah, it should stay as it is.

Syncano/SyncanoException.m
lifcio: I'm not sure I fully understand. They way you suggest it (if I understood correctly) would be for the two functions to behave differently. So assuming both functions are called with the code 1234, `exceptionWithCode:` would do A and `exceptionWithCode:  reason:` would do B?
That's totally inconsistent in my opinion and should never be the case. Do you have any use case in mind?

PR #10, opened 2015-08-14
Create SCDataObject from dictionary

lifcio:
Allows creating SCDataObject using dictionary. Uses SCParseManager
behind the scenes and hides it from the user.

Helpful when working with channels - when new objects are created,
channel notification contains payload dictionary, which is an object.
This method allows easy object creation using received payload.


lifcio:
@jaskooo Please take a look if this is the best way to solve it.

lifcio:
Merging.

PR #12, opened 2015-09-08
run webhook method refactor

jaskooo:


jaskooo:
tests failed in circleci because they running tests using older version of XCode. They anounced upgrade soon.

lifcio:
LGTM. We should also add in next version one more method, where you can pass params (they will be available inside a codebox).

PR #19, opened 2015-09-19
SCCompoundPredicate added

jaskooo:


syncano-ios/SCCompoundPredicate.m
```diff
@@ -0,0 +1,64 @@
...
+    [self.predicates addObject:predicate];
+}
+
+-(NSString *)queryRepresentation {
+    NSMutableString* query = [NSMutableString stringWithString:@"{"];
+    
+    int i=0;
+    for(id<SCPredicateProtocol> predicate in self.predicates) {
+        if(i!=0)
+            [query appendString:@","];
```
lifcio: Can you add curly braces (even for one line ifs) and a space between if and condition? 
i.e.:
```objc
if (i!=0) {
   [query appendString:@","];
}
```

syncano-ios/SCCompoundPredicate.m
```diff
@@ -0,0 +1,64 @@
...
+    }
+    return self;
+}
+
++ (instancetype)compoundPredicateWithPredicates:(NSArray *)predicates {
+    return [[self alloc] initWithPredicates:predicates];
+}
+
+- (void)addPredicate:(id<SCPredicateProtocol>)predicate {
+    if (!predicate) {
```
lifcio: Shouldn't we also check here if the predicate conforms to the protocol?

lifcio:
Other than the comments, looks good!

PR #20, opened 2015-09-24
File upload

jaskooo:


syncano-ios/SCDataObject+Properties.m
```diff
@@ -0,0 +1,50 @@
...
+        const char *propName = property_getName(property);
+        if(propName) {
+            NSString *propertyName = [NSString stringWithUTF8String:propName];
+            NSString* propertyAttributes = [NSString stringWithUTF8String:property_getAttributes(property)];
+            NSArray* splitPropertyAttributes = [propertyAttributes componentsSeparatedByString:@"\""];
+            if ([splitPropertyAttributes count] >= 2)
+            {
+                NSString *className = [splitPropertyAttributes objectAtIndex:1];
+                results[propertyName] = className;
+                NSLog(@"Property: %@ - Class: %@",propertyName, className);
```
lifcio: Change to some kind of debug log?

syncano-ios/SCFile.m
```diff
@@ -9,9 +9,29 @@
...
+
+@interface SCFile ()
+@property (nonatomic,retain) NSData *data;
+@property (nonatomic) BOOL needsToBeUploaded;
+@end
 
 @implementation SCFile
 
++ (instancetype)fileWithaData:(NSData *)data {
+    return [[SCFile alloc] initWithData:data];;
```
lifcio: Double `;;`

lifcio: Should be `[self alloc]`

syncano-ios/SCDataObject+Properties.m
jaskooo: I left this one by mistake while i was testing. I think itvshould be removed from here.

syncano-ios/SCFile.m
```diff
@@ -26,6 +46,32 @@ + (NSDictionary *)JSONKeyPathsByPropertyKey {
...
+    [self saveAsPropertyWithName:name ofDataObject:dataObject usingAPIClient:[Syncano sharedAPIClient] withCompletion:completion];
+}
+
+- (void)saveAsPropertyWithName:(NSString *)name ofDataObject:(SCDataObject *)dataObject toSyncano:(Syncano *)syncano withCompletion:(SCCompletionBlock)completion {
+    [self saveAsPropertyWithName:name ofDataObject:dataObject usingAPIClient:syncano.apiClient withCompletion:completion];
+}
+
+- (void)saveAsPropertyWithName:(NSString *)name ofDataObject:(SCDataObject *)dataObject usingAPIClient:(SCAPIClient *)apiClient withCompletion:(SCCompletionBlock)completion {
+    if (self.needsToBeUploaded) {
+        [apiClient postUploadTaskWithPath:dataObject.path propertyName:name fileData:self.data completion:^(NSURLSessionDataTask *task, id responseObject, NSError *error) {
```
lifcio: How does it handle no data (`data == nil`)? It would be the case I think if we get the object with a file inside.

syncano-ios/SCFile.m
```diff
@@ -26,6 +46,32 @@ + (NSDictionary *)JSONKeyPathsByPropertyKey {
...
+        }];
+    } else {
+        if (completion) {
+            completion(nil);
+        }
+    }
+}
+
 - (void)fetchInBackgroundWithCompletion:(SCFileFetchCompletionBlock)completion {
     [SCAPIClient downloadFileFromURL:self.fileURL withCompletion:^(id responseObject, NSError *error) {
```
lifcio: Shouldn't fetching also save the file in the `data` variable? At least as an option? This way I wouldn't have to fetch it form the Internet every time I want to access it. I know that holding in memory 128mb files isn't a way to go, but if I have e.g. 100 objects, each with e.g. a JSON 1kb file - holding them inside SCFile as NSData might be the easiest way to go.

syncano-ios/SCFile.m
jaskooo: i think if `data == nil' then `self.needsToBeUploaded` will be `NO`

lifcio: When new file is allocated it's setting `self.needsToBeUploaded` to `YES`. Then it's set to `NO` only after `save`. So probably we will have not needed call going to Syncano with an empty `data` field.

jaskooo: Right. ;)

PR #30, opened 2015-10-23
Framework

jaskooo:
Changed xcode target from static lib to framework to reduce problems with builds


syncano-ios.xcodeproj/project.pbxproj
```diff
@@ -489,32 +494,52 @@
 				C6C26CD1570D29893B2F95D7 /* libPods-syncano4-iosTests.a */,
```
lifcio: Something old? Can be removed?

syncano-ios/Syncano.h
```diff
@@ -23,6 +25,14 @@
 #import "SCChannel.h"
 #import "SCFile.h"
 
+//! Project version number for Syncano.
+FOUNDATION_EXPORT double SyncanoVersionNumber;
+
+//! Project version string for Syncano.
+FOUNDATION_EXPORT const unsigned char SyncanoVersionString[];
+
+// In this header, you should import all the public headers of your framework using statements like #import <Syncano/PublicHeader.h>
```
lifcio: Looking at the comment - why are we doing `#import "SCFile.h"` instead? Either comment or our import is wrong.

lifcio:
Not sure what happened here exactly - did we change the name of our Pod? Is it `Syncano` instead of `syncano-ios` now?

syncano-ios/Syncano.h
jaskooo: No. This change is only in xcode project. I have changed target from static lib to framework. It should not affect Pods

syncano-ios.xcodeproj/project.pbxproj
jaskooo: Sure :)

PR #31, opened 2015-11-04
Offline support - saving Data Objects

jaskooo:


syncano-ios/SCConstants.m
```diff
@@ -95,12 +95,10 @@ + (NSString *)DB_PATH {
 }
 
 + (NSString *)createTableSQLStatement {
-    NSString *objectsTableSchema = @"CREATE TABLE IF NOT EXISTS %@ ("
-    @"Syncano TEXT PRIMARY KEY, "
+    NSString *objectsTableSchema = @"CREATE TABLE IF NOT EXISTS SyncanoDataObjects ("
```
lifcio: Same as other PR, we should try to move table name to an external variable.

syncano-ios/SCDataObject+LocalStorage.m
```diff
@@ -0,0 +1,29 @@
...
+#import "SCDataObject+LocalStorage.h"
+#import "SCParseManager+SCDataObject.h"
+#import "Syncano.h"
+#import "SCLocalStore.h"
+
+@implementation SCDataObject (LocalStorage)
+
+#pragma mark - Saving -
+- (void)saveToLocalStorageWithCompletion:(SCCompletionBlock)completion {
+    [[Syncano localStore] saveDataObject:self withCompletionBlock:completion];
```
lifcio: Same as in other PR - if inside of `SCLocalStore` `if ([_db open]) {` will bring false, `completion` block will never be called.

lifcio:
Similar issues to the ones before - fixing them there should fix them here as well. Other than that looks good.

lifcio:
Same as in https://github.com/Syncano/syncano-ios/pull/32 , other than that looks good

PR #32, opened 2015-11-05
Offline support json serializer for do

jaskooo:


syncano-ios/SCConstants.m
```diff
@@ -95,12 +95,10 @@ + (NSString *)DB_PATH {
 }
 
 + (NSString *)createTableSQLStatement {
-    NSString *objectsTableSchema = @"CREATE TABLE IF NOT EXISTS %@ ("
-    @"Syncano TEXT PRIMARY KEY, "
+    NSString *objectsTableSchema = @"CREATE TABLE IF NOT EXISTS SyncanoDataObjects ("
```
lifcio: Is this the only place where you use this table name? I think previous approach made sense, with using a variable here, that would hold table name inside.

syncano-ios/SCDataObject+LocalStorage.m
```diff
@@ -0,0 +1,29 @@
...
+
+#pragma mark - Saving -
+- (void)saveToLocalStorageWithCompletion:(SCCompletionBlock)completion {
+    [[Syncano localStore] saveDataObject:self withCompletionBlock:completion];
+}
+
+#pragma mark - Helpers -
+- (void)generateInsertQueryWithCompletion:(void(^)(NSError *error, NSString* query))completion {
+    NSError *error;
+    NSString *insertQuery = [NSString stringWithFormat:@"INSERT INTO SyncanoDataObjects (className, objectId,json) VALUES ('%@',%@,'%@');",NSStringFromClass([self class]),self.objectId,[[SCParseManager sharedSCParseManager] JSONSerializedDictionaryFromDataObject:self error:&error]];
```
lifcio: I see we use table name explicitly here as well. If we ever changed it, we would need to be sure to change it everywhere. Better to store i as a variable.

syncano-ios/SCLocalStore.m
```diff
@@ -32,7 +33,19 @@ - (void)initializeDB {
 }
 
 - (void)saveDataObject:(SCDataObject *)dataObject withCompletionBlock:(SCCompletionBlock)completionBlock {
-    
+    if ([_db open]) {
```
lifcio: Why not use `self.db` here?

syncano-ios/SCLocalStore.m
```diff
@@ -32,7 +33,19 @@ - (void)initializeDB {
...
+                [_db beginTransaction];
+                [_db executeUpdate:query];
+                [_db  commit];
+            }
+            [_db close];
+            if (completionBlock) {
+                completionBlock(error);
+            }
+        }];
+    }
```
lifcio: What happens if `[_db open]` returns false? Completion block will never be called - probably should still be called, just with an error.

syncano-ios/Syncano.m
```diff
@@ -12,13 +12,13 @@
 #import "SCAPIClient+Class.h"
 #import "SCLocalStore.h"
 
+static SCLocalStore *localStore;
```
lifcio: Since it's an internal variable, maybe use underscore at the beginning, e.g. `__localStore`.

lifcio:
Why moved all `SCClassRegisterItem` to main class? Not saying it's bad, just curious :)

jaskooo:
SCClassregisterItem was moved to main class because it is common for two different categories.
I want to move registering classes to a separate class though.

lifcio:
Makes sense.

Overall comment after the changes - looks good, but I don't understand why in some places you `- (void)handleError:(NSError *)error forCompletionBlock:(SCCompletionBlock)completionBlock` and in some you don't. Is there a logic behind it?

E.g. [here](https://github.com/Syncano/syncano-ios/pull/32/files#diff-404c07fc5bc7ca113800decb6ac57d31R40) you use the method, but [here](https://github.com/Syncano/syncano-ios/pull/32/files#diff-404c07fc5bc7ca113800decb6ac57d31R106) you don't. Should all of them use it?

PR #37, opened 2015-11-12
Enumerate pages

jaskooo:


syncano-ios/SCAPIClient.m
```diff
@@ -30,7 +30,8 @@ - (instancetype)initWithBaseURL:(NSURL *)url {
     self = [super initWithBaseURL:url];
     if (self) {
         self.requestSerializer = [AFJSONRequestSerializer serializer];
-        self.securityPolicy.allowInvalidCertificates = NO;
+        self.securityPolicy.allowInvalidCertificates = YES;
```
lifcio: It was changed to NO, cause for some reason checking the cert didn't allow to connect to Syncano. Can you confirm it works now?

lifcio:
LGTM, I like the solution!

syncano-ios/SCAPIClient.m
lifcio: NVM, looks good, I read the property name wrong. All is fine, move along.

PR #44, opened 2015-12-09
LIB-191 added SCRequest an SCRequestCacheManager

jaskooo:


syncano-ios/SCRequest.m
```diff
@@ -0,0 +1,69 @@
...
+//
+
+#import "SCRequest.h"
+
+static NSString * const kRequestMethodGET = @"GET";
+static NSString * const kRequestMethodPOST = @"POST";
+static NSString * const kRequestMethodPATCH = @"PATCH";
+static NSString * const kRequestMethodDELETE = @"DELETE";
+static NSString * const kRequestMethodPUT = @"PUT";
+static NSString * const kRequestMethodUnknown = @"UNKNOWN";
```
lifcio: We should keep it consistent - enum and string. It should be in both places either `Unknown` or `Undefined`

syncano-ios/SCRequestCacheManager.m
```diff
@@ -0,0 +1,31 @@
...
+#import "SCRequest.h"
+
+@interface SCRequestCacheManager ()
+@property (nonatomic,retain) NSMutableDictionary *requestsStore;
+@end
+
+@implementation SCRequestCacheManager
+- (void)enqueueRequest:(SCRequest *)request {
+    if (!self.requestsStore) {
+        self.requestsStore = [NSMutableDictionary new];
```
lifcio: It's better to write a custom getter for `requestStore` and there check if exists or not (not a big deal here, but will make it easier to maintain).

lifcio:
Other than mentioned issues - looks good!

syncano-ios/SCRequestCacheManager.m
```diff
@@ -0,0 +1,36 @@
...
+@property (nonatomic,retain) NSMutableDictionary *requestsStore;
+@end
+
+@implementation SCRequestCacheManager
+
+- (NSMutableDictionary *)requestsStore {
+    if (!_requestsStore) {
+        _requestsStore = [NSMutableDictionary new];
+    }
+    return _requestsStore;
```
lifcio: Perfect! :)

lifcio:
Good to merge now.

PR #45, opened 2015-12-11
add requests to queue

jaskooo:
now all requests are added to queue


dancio:
@lifcio any thoughts?

PR #48, opened 2015-12-28
Lib 349 run queue

jaskooo:


syncano-ios/SCAPIClient.m
```diff
@@ -72,11 +80,100 @@ - (void)authorizeRequest {
...
+
+- (void)runQueue {
+    if (self.requestQueueState == SCAPIClientRequestQueueStateRunning) {
+        return;
+    }
+    [self dequeueNextRequest];
+}
+
+- (void)dequeueNextRequest {
+    if (self.requestQueue.hasRequests) {
```
lifcio: Nice, better than `.count > 0` :)

syncano-ios/SCAPIClient.m
```diff
@@ -72,11 +80,100 @@ - (void)authorizeRequest {
...
+#pragma mark  - Dequeue -
+
+- (void)runQueue {
+    if (self.requestQueueState == SCAPIClientRequestQueueStateRunning) {
+        return;
+    }
+    [self dequeueNextRequest];
+}
+
+- (void)dequeueNextRequest {
```
lifcio: Should we think about threading here? Should we lock/unlock this function (or `runQueue`) so there's no more than one entrance inside? Or if someone uses our API directly (without creating own threads) it should work fine?

syncano-ios/SCAPIClient.m
```diff
@@ -72,11 +80,100 @@ - (void)authorizeRequest {
...
+    if (self.requestQueue.hasRequests) {
+        self.requestQueueState = SCAPIClientRequestQueueStateRunning;
+        SCRequest *request = self.requestQueue.nextRequest;
+        [self runRequest:request];
+        [self.requestQueue removeRequest:request];
+        [self dequeueNextRequest];
```
lifcio: If you run `dequeueNextRequest` here directly, it's gonna loop until whole queue is empty, without waiting for previous requests to finish - was that your intention? In `runQueue` you check `state` first. We can discuss it outside of GH and just post result of discussion here :)

syncano-ios/SCAPIClient.m
```diff
@@ -72,11 +80,100 @@ - (void)authorizeRequest {
...
+    }
+    [self dequeueNextRequest];
+}
+
+- (void)dequeueNextRequest {
+    if (self.requestQueue.hasRequests) {
+        self.requestQueueState = SCAPIClientRequestQueueStateRunning;
+        SCRequest *request = self.requestQueue.nextRequest;
+        [self runRequest:request];
+        [self.requestQueue removeRequest:request];
```
lifcio: Usually, when you implement a queue - you add `dequeue` method, that returns object from the beginning and removes it.
So instead of 
```objc
SCRequest *request = self.requestQueue.nextRequest;
 [self runRequest:request];
 [self.requestQueue removeRequest:request];
```
you could have

```objc
SCRequest *request = [self.requestQueue dequeue];
[self runRequest:request];
```
This way, it acts more as a queue and not an array :)

syncano-iosTests/SCAPIClientSpec.m
```diff
@@ -172,57 +174,57 @@
...
+//                _responseobject = responseObject;
+//                _blockFinished = YES;
+//            }];
+//            [[expectFutureValue(theValue(_blockFinished)) shouldEventually] beYes];
+//            [[_error shouldEventually] beNil];
+//            [[task.currentRequest.HTTPMethod should] equal:@"GET"];
+//            [[task.currentRequest.URL should] equal:url];
+//            [[task.currentRequest.allHTTPHeaderFields[@"X-API-KEY"] should] equal:@"API-KEY"];
+//            [[_responseobject shouldEventually] beNonNil];
+//        });
```
lifcio: Make sure at some point this code is uncommented or removed ;)

syncano-ios/SCRequestQueue.m
```diff
@@ -24,46 +24,54 @@ - (instancetype)initWithIdentifier:(NSString *)identifier {
     return self;
 }
 
-- (NSMutableDictionary *)requestsStore {
+- (BOOL)hasRequests {
+    return self.requestsStore.count > 0;
+}
+
+- (SCRequest *)nextRequest {
```
lifcio: Rather than using `next` I would use `first`, will be easier to understand that it just returns first element in the queue (Java or C# use `peek`, C++ std uses `front`, so feel free to use any of these as well, though `first` will match Foundation naming better)

lifcio:
Other than comments above, looks good to me. So when we take of them, can be merged :)

PR #54, opened 2016-01-15
Lib 428 revision mismatch notify

jaskooo:


syncano-ios/NSError+RevisionMismatch.m
```diff
@@ -0,0 +1,31 @@
...
+#import "NSError+RevisionMismatch.h"
+#import "SCConstants.h"
+
+@implementation NSError (RevisionMismatch)
+- (void)checkIfMismatchOccuredWithCompletion:(SCDataObjectRevisionMismatchCompletionBlock)completionBlock {
+    if (self.userInfo[kSyncanoRepsonseErrorKey]) {
+        NSDictionary *syncanoErroInfo = self.userInfo[kSyncanoRepsonseErrorKey];
+        BOOL mismatched = NO;
+        NSString *description;
+        if (syncanoErroInfo[@"expected_revision"]) {
```
lifcio: Can we make it a constant? I see there were some changes with errors on the Core side, just in case it happens again - it would be better to have all error strings in one place.

syncano-ios/NSError+RevisionMismatch.m
```diff
@@ -0,0 +1,31 @@
...
+            mismatched = YES;
+            description = syncanoErroInfo[@"expected_revision"];
```
lifcio: Instead of accessing the dictionary twice, you could get the description only once. E.g.:
```objc
        NSString *expectedRevisionDescription = syncanoErroInfo[@"expected_revision"];
        if (expectedRevisionDescription) {
            mismatched = YES;
        }
        if (completionBlock) {
            completionBlock(mismatched,expectedRevisionDescription);
        }
```

or even rewrite the whole thing, to remove code repetition:
```objc
- (void)checkIfMismatchOccuredWithCompletion:(SCDataObjectRevisionMismatchCompletionBlock)completionBlock {
    NSDictionary *syncanoErroInfo = self.userInfo[kSyncanoRepsonseErrorKey];
    NSString *expectedRevisionDescription = syncanoErroInfo[@"expected_revision"];
    BOOL mismatched = NO;
    if (expectedRevisionDescription) {
        mismatched = YES;
    }
    if (completionBlock) {
        completionBlock(mismatched,expectedRevisionDescription);
    }
}
```
Didn't check it but should work too.

lifcio:
Other than comments - looks good! I like the idea of adding a function, that passes current revision as `expected_revision` - so user doesn't need to set it explicitly - and then has two ways of saving objects - with checking, and without checking.
Though what do we do now with future `saveEventually`? Probably questions for later :)

syncano-ios/NSError+RevisionMismatch.m
jaskooo: It is working. Thanks

PR #55, opened 2016-01-21
Lib 442 integration tests env

jaskooo:


lifcio:
Cool, looks good! :)

One question though - if we're using environment settings to store API Key, maybe we should do the same for the instance name as well?

jaskooo:
Good point. will move INSTANCE_NAME there

PR #57, opened 2016-01-25
[LIB-183] Better string filtering

trapi:


lifcio:
LGTM, clean and simple :)

PR #58, opened 2016-01-26
LIB-170 data views support

trapi:


syncano-ios/SCParseManager+SCDataObject.h
```diff
@@ -52,6 +52,21 @@
...
+ *  @param objectClass Class of object to parse for
+ *  @param JSONObject  serialized JSON object from API response
+ *  @param relations   Key is class name, value is Dictionary of relations for the class @see relationsForClass. Pass an empty dictionary here
+ *  @param error       Error object
+ *
+ *  @return parsed SCDataObject
+ */
+- (id)parsedObjectOfClass:(__unsafe_unretained Class)objectClass
+           fromJSONObject:(id)JSONObject
+           withRelationsCache:(NSMutableDictionary<NSString*,NSDictionary*> *)relationsCache
```
trapi: I'm not very happy with this NSMutableDictionary but since we use a singleton and IMO looking for relations each time in a loop is too much time consuming.

trapi:
3 Parts of this task here:
1. PleaseForView in SCDataObject
2. SCDataObject "isView"
3. Parsing references (expand options in view)

lifcio:
Started reviewing, will finish tomorrow, so please don't merge before. But looks good so far.

syncano-ios/SCDataObject.m
```diff
@@ -51,13 +56,26 @@ + (void)registerClass {
 }
 
 + (SCPlease *)please {
+    if([self viewNameForAPI] != nil)
+        return [self pleaseForView:[self viewNameForAPI]];
```
lifcio: You don't use curly braces after `if`? Swift will teach you to use them ;)

syncano-ios/SCParseManager+SCDataObject.h
lifcio: Hmm, can you guide through it? I don't really see any big improvement for such a huge change. I don't know if I would call it cache - the dictionary isn't cached anywhere and every time we enter this method we start building it from scratch.
The only improvement I see is that in
```objc
- (NSDictionary *)relationsForClass:(__unsafe_unretained Class)class fromCache:(NSMutableDictionary<NSString*,NSDictionary*> *)relationsCache
```
you get `relations = [self relationsForClass:class];` only once per class. But this doesn't change much from practical point of view - it will be useful only if a class points multiple times to other class. E.g. Book has multiple references to Author.
Is that the whole improvement, or did I miss something?

lifcio:
I like the idea of adding new pleaseForView methods, and being able to use views with known convention of `SCPlease` :+1: Good job!

After you answer my question about the `relationsCache` we will be able to merge - everything else looks good!

syncano-ios/SCParseManager+SCDataObject.h
trapi: Maybe you're right that it's not worth it. You missed one thing:
We're looking at develop branch. 
First of all, we all agree that `relationsForClass:` is quite slow because it loops trough all properties in a class, do we?. In `SCParseManager+SCDataObject.m` `relationsForClass:`is used by `resolveRelationsToObject:withJSONObject:`which is used by `parsedObjectOfClass:fromJSONObject:`. `parsedObjectOfClass:fromJSONObject:`is used by `parsedObjectsOfClass:fromJSONObject:`.
This `parsedObjectsOfClass:fromJSONObject:` is used by every data fetch from Syncano.
Let assume that you fetch 100 objects (Book) with 30 properties and 1 relation to class with 6 properties (Author) (U will have a lot of properties because these include 'created_at', 'owner' etc.). So you invoke `relationsForClass` 100 times for Book and 100 times for Author having the same results. It's not a lot but it can be easily changed to 2 invokes which is 100 time less in this example;)

That's why this cache is needed. I did it this way because we have singleton + `relationsForClass` results may vary in time (mostly after registration of new class). What's more this all may be used in a concurrent code.
@jaskooo: Any thoughts on this?

syncano-ios/SCDataObject.m
trapi: No I do not. If you have different convention in this project I will stick to it and use them ;)

syncano-ios/SCParseManager+SCDataObject.h
jaskooo: I think we should think about a best solution. Yours looks good for me and i thinks @lifcio dont like only the name of this argument. if you change `cache` into something else he will be happy with this :)

There is a second thing with those changes. On `offline-support` branch i have already changed a structure and a way that classes are `registeres` please reffer to this merge: https://github.com/Syncano/syncano-ios/commit/77a247968c9724bd9c84c681b047755cdfacb59a
maybe cherry pick o this will be enough? :smile:

syncano-ios/SCDataObject.m
lifcio: We didn't really make an official decision on that, but I see @jaskooo adds braces for one line `ifs` as well, so might be better to be consistent here.

jaskooo: Yes. Use the braces.

syncano-ios/SCParseManager+SCDataObject.h
lifcio: Ahh. Now I see. So I did miss it. I was analyzing only the method I commented on - `parsedObjectOfClass` which isn't improved if used as explained in comments - by always passing an empty dictionary.
But I see that `parsedObject*s*OfClass` create this dictionary once and pass the same object to subsequent calls of `parsedObject` The improvement was just in a different place than I looked for :)

And yes, what @jaskooo said - I think `cache` name is misleading, here. Maybe just describe what's inside? 
I suggest `relationsForClass` -- then we will have also more readable code. [This line](https://github.com/Syncano/syncano-ios/pull/58/files#diff-1c99c2ef6f67208cb474dc899ab40f72R143) could become:
```objc
NSDictionary* relations = relationsForClass[className];
```
etc.

trapi:
I refreshed this branch with offline support and instead of this "relationsCache" I created RelationsForClass dict in SCRegisterManager. Please review this and let's go with this merge :)

lifcio:
lgtm

PR #59, opened 2016-01-27
[LIB-203] Add support for class JOIN

trapi:


syncano-ios/SCPredicate.h
```diff
@@ -43,6 +43,8 @@
 
 + (SCPredicate *)whereKey:(NSString *)key inArray:(NSArray *)array;
 
++ (SCPredicate *)whereKey:(NSString *)key satisfiesPredicate:(id<SCPredicateProtocol>)predicate;
```
trapi: I'm not sure if this name is 100% accurate. What do you think?

jaskooo: It is good! Leave it like this

jaskooo:
LGTM :) Very nice change!

syncano-ios/SCCompoundPredicate.m
```diff
@@ -43,23 +43,28 @@ - (void)addPredicate:(id<SCPredicateProtocol>)predicate {
     [self.predicates addObject:predicate];
 }
 
--(NSString *)queryRepresentation {
-    NSMutableString* query = [NSMutableString stringWithString:@"{"];
-    
-    int i=0;
+- (NSDictionary<NSString*,NSDictionary*> *)rawPredicate {
```
lifcio: That's great! :+1: 

We should try to update returned containers to new syntax everywhere where possible - it will make working with our library in Swift much easier. But that's of course outside of the scope of this ticket :)

jaskooo: :+1:

PR #60, opened 2016-01-27
Lib 446 offline integration test

jaskooo:


SyncanoIntegrationTests/OfflineStorageSpec.m
```diff
@@ -0,0 +1,63 @@
...
+                    if(error){
+                        NSLog(@"error: %@",error);
+                    }
+                }];
+            }];
+            [[expectFutureValue(theValue(_blockFinished)) shouldEventuallyBeforeTimingOutAfter(10.0)] beYes];
+            [[_error shouldEventually] beNil];
+            [[_book shouldEventually] beNonNil];
+            [[_storedBook shouldEventually] beNonNil];
+            [[_bookId shouldEventually] equal:_storedBook.objectId];
```
trapi: Shouldn't we check all fields on Book object? Maybe the code works for ids but not for other custom field (f.e. text)?

jaskooo: Why not. Good idea :)

SyncanoIntegrationTests/OfflineStorageSpec.m
```diff
@@ -0,0 +1,63 @@
...
+            __block Book *_book;
+            __block NSNumber *_bookId;
+            __block Book *_storedBook;
+            
+            //Getting books from API
+            [[Book please] giveMeDataObjectsWithCompletion:^(NSArray *objects, NSError *error) {
+                _book = [objects firstObject];
+                _bookId = _book.objectId;
+                // Savig first of them
+                [_book saveToLocalStorageWithCompletion:^(NSError *error) {
```
trapi: Maybe I missed it but this test should clear the local store before writing something otherwise it might be in conflict with other (future) tests.

jaskooo: You are right. Will handle that. I think i will have to create a method for this feature.

lifcio:
Other than what Kuba mentioned LGTM

PR #62, opened 2016-01-29
Default permissions hot fix

lifcio:


lifcio:
We need something better - ideally we shouldn't set any defaults in our library and depend on default values from the server - in this case it's just for a quick hotfix, and I will start working on a better solution.

PR #65, opened 2016-01-30
[LIB-475] Do not send object permissions if they were not set

lifcio:


lifcio:
Hotfix I already pushed to cocoapods temporarily solves the problem, but we cannot change default permissions on the library side whenever they change in the backend (better to set defaults in one place - backend, than in 10 libraries).

@trapi @jaskooo What do you think? 
My only concern would be not adding in the Mantle transformer `Default` to `states` as there is no value for it - backend will always return one of `none`, `full`, `write`, `read`.
In the reverse transformer, when it doesn't find it in `states` it returns `nil` - so it's quite convenient.

syncano-ios/SCConstants.h
```diff
@@ -49,6 +49,7 @@ extern NSString * const kSCChannelNotificationMessageActionUpdate;
 extern NSString * const kSCChannelNotificationMessageActionDelete;
 
 typedef NS_ENUM(NSUInteger, SCDataObjectPermissionType) {
+    SCDataObjectPermissionTypeDefault,
```
trapi: I have kind of fear when I add new enum at the begging of an existing one. Is it only me?

syncano-ios/SCDataObject.m
```diff
@@ -62,14 +62,6 @@ + (instancetype)objectFromDictionary:(NSDictionary *)dictionary {
     return [[SCParseManager sharedSCParseManager] parsedObjectOfClass:self fromJSONObject:dictionary];
 }
 
-- (id)init {
-    self = [super init];
-    if (self) {
-        self.owner_permissions = SCDataObjectPermissionTypeFull;
```
trapi: I don't like to really on default values of variables. IMO change to explicit value of SCDataObjectPermissionTypeDefault would be enough.

trapi:
I think it's OK and default permissions should be left for server to decide.

syncano-ios/SCConstants.h
lifcio: True, some other code might depend on the integer value of this enum. So, as a fix - add it at the end - and then maybe also add explicit integer values?

syncano-ios/SCDataObject.m
lifcio: Ok, agree. If we don't set it to explicit value, adding a new enum value at the beginning will change behavior completely :+1:

lifcio:
Scope of this ticket got a big bigger.
* I needed to refresh permissions after saving object (so they are not `NotSet` after saving). If I added that in `updateObjectAfterSaveWithDataFromJSONObject` - it would require some code repetition. 
* And I noticed when testing, we never actually were transforming dates - they were saved as strings inside NSDate* variable - so another value transformer was added
* Now the best choice to update object just seemed to be through `SCParseManager` and `fillObject` method.

My one concern here is with using `NSDateFormatter` in value transformer - it used to be quite slow. That's why in v3 branch I used - https://github.com/Syncano/syncano-ios/blob/v3/Syncano/Utilities/SyncanoDateFormatter.m#L50 - it was 2x-10x faster than `NSDateFormatter`, which made a huge difference when downloading thousands of objects. 
For now I say we keep using `NSDateFormatter`, but prepare some performance tests and keep in mind possible different solution(s).

trapi:
"And I noticed when testing, we never actually were transforming dates - they were saved as strings inside NSDate* variable - so another value transformer was added" True. It should be changed for all fields of a type NSData not only for "created_at" and "updated_at".

lifcio:
True. Thanks! I will make a new ticket for it, as this got bigger than I wanted it to be :)

PR #67, opened 2016-02-02
[Lib-379] Save SCFile content to disk with progress

trapi:


trapi:
@lifcio Any comments?

SyncanoIntegrationTests/FileFetchSpec.m
```diff
@@ -0,0 +1,80 @@
...
+        __block int64_t _bytesWritten = 0;
+        __block Book* _book;
+        [[Book please] giveMeDataObjectsWithPredicate:[SCPredicate whereKey:@"id" isEqualToNumber:@(8)] parameters:nil completion:^(NSArray *objects, NSError *error) {
+            if(error) {
+                _errorPlease = error;
+                _blockFinished = YES;
+                return;
+            }
+            
+            _book = objects[0];
```
lifcio: Use `.firstObject` unless you need it to crash if the array is empty (but then, better to just test if array is empty)

SyncanoIntegrationTests/FileFetchSpec.m
```diff
@@ -0,0 +1,80 @@
...
+    it(@"download a file to disk", ^{
+        NSURL *storePathDocuments = [[NSFileManager defaultManager] URLForDirectory:NSDocumentDirectory inDomain:NSUserDomainMask appropriateForURL:nil create:NO error:nil];
+        NSURL *storePath = [storePathDocuments URLByAppendingPathComponent:@"saabManual.pdf"];
+        
+        __block BOOL _blockFinished;
+        __block NSError* _errorPlease;
+        __block NSError* _errorFile;
+        __block int64_t _bytesWritten = 0;
+        __block Book* _book;
+        [[Book please] giveMeDataObjectsWithPredicate:[SCPredicate whereKey:@"id" isEqualToNumber:@(8)] parameters:nil completion:^(NSArray *objects, NSError *error) {
```
lifcio: I don't have any strong opinion here, just thinking out loud - wouldn't it be better to send the file each time? It will make the test longer (so we may choose a smaller file), but we will avoid potential issue of file disappearing from the server (if it happens, tests will need to be updated with new object id). @jaskooo ?

syncano-ios/SCFile.m
```diff
@@ -83,10 +83,19 @@ - (void)fetchInBackgroundWithCompletion:(SCFileFetchCompletionBlock)completion {
...
             if (completion) {
-                completion(self.data,nil);
+                completion(responseObject,nil);
             }
         }
     }];
 }
 
+- (NSURLSessionDownloadTask *)fetchToFileInBackgroundWithProgress:(SCFileDownloadProgressCompletionBlock)progress completion:(SCFileFetchToDiskCompletionBlock)completion {
+    return [self fetchToFileInBackground:self.storeURL withProgress:progress completion:completion];
```
lifcio: Should we handle situation when `storeURL == nil`? We could grab filename from `fileURL` and save it under this name in `NSTemporaryDirectory` ?

lifcio:
Other than above - looks good!

SyncanoIntegrationTests/FileFetchSpec.m
trapi: Good point.

SyncanoIntegrationTests/FileFetchSpec.m
trapi: I think that this is a fundamental decision for all integration tests we have. We can:
1. Build schema and put test data for each test and clean after it. We cannot do everything from the library, so sometimes raw HTTP call would be neede
2. Depend on existing schema/test data - this is our current direction - for all tests we assume existence of some data.

The problem with testing file downloading is even more complicated. Did you notice this line: `if([[NSFileManager defaultManager] isWritableFileAtPath:storePathDocuments.path])`. This returns false on cirecleci and xctool in general :(

syncano-ios/SCFile.m
trapi: Nice feature. I will implement it.

SyncanoIntegrationTests/FileFetchSpec.m
lifcio: Ugh, ok. I'm good with how it is now and agree if we change it it should be applied to all integrations tests. Not sure if we can get feedback from @jaskooo this week, so let's leave it as it is for now and figure it out later.

syncano-ios/SCFile.m
lifcio: @23doors : Can we assume file names will be always unique for different files? (is it possible two files stored on Syncano will have the same filename?)

23doors: @lifcio sorry missed that. Yes, I think it is safe to assume. Chances of collision across all of the clients are very, very small. And for 1 app we can just ignore those chances altogether.

PR #68, opened 2016-02-04
[LIB-167] increment decrement fields

trapi:


trapi:
I'm not sure if we need all this methods for increment - especially with this revision mismatch block.. @jaskooo can you comment on this? It's based on updateValue implementation.

SyncanoIntegrationTests/DataObjectManipulationsSpec.m
```diff
@@ -0,0 +1,104 @@
...
+
+SPEC_BEGIN(DataObjectManipulationsSpec)
+
+describe(@"Data object manipulations", ^{
+    
+    NSDictionary *environment = [[NSProcessInfo processInfo] environment];
+    NSString *apiKey = environment[@"API_KEY"];
+    NSString *instanceName = environment[@"INSTANCE_NAME"];
+    
+    [Syncano sharedInstanceWithApiKey:apiKey instanceName:instanceName];
```
lifcio: Just a question: In LIB-379 you initialized `Syncano` inside `beforeAll` block and here it's outside of it - what is the main difference between doing it one way or another?

SyncanoIntegrationTests/DataObjectManipulationsSpec.m
```diff
@@ -0,0 +1,104 @@
...
+            [[book.lovers should] equal:expectedLoversNb];
+            
+        });
+        
+        it(@"should increment many values", ^{
+            __block BOOL _blockFinished;
+            __block NSError *_error;
+            NSNumber* expectedLoversNb = @(book.lovers.integerValue - 1);
+            NSNumber* expectedReadersNb = @(book.readers.integerValue + 12);
+            [book incrementValues:@{@"lovers":@(-1),@"readers":@12} withCompletion:^(NSError *error) {
```
lifcio: For consistency with previous example use @(12)

SyncanoIntegrationTests/DataObjectManipulationsSpec.m
```diff
@@ -0,0 +1,104 @@
...
+            [[book.lovers should] equal:expectedLoversNb];
+            [[book.readers should] equal:expectedReadersNb];
+        });
+    });
+    
+    context(@"Errors", ^{
+        it(@"should error on unknown property", ^{
+            __block BOOL _blockFinished;
+            __block NSError *_error;
+            [book incrementKey:@"haters" by:@1 withCompletion:^(NSError *error) {
```
lifcio: @(1)

syncano-ios/SCDataObject+Increment.m
```diff
@@ -0,0 +1,91 @@
...
+#import "SCDataObject+Increment.h"
+#import "SCAPIClient.h"
+#import "NSError+RevisionMismatch.h"
+
+@implementation SCDataObject (Increment)
+
+- (NSError*)errorForUnknownProperty:(NSString*)propertyName {
+    NSDictionary *userInfo = @{
+                               NSLocalizedDescriptionKey: [NSString stringWithFormat:NSLocalizedString(@"Property %@ does not exist", @""),propertyName],
+                               NSLocalizedFailureReasonErrorKey: NSLocalizedString(@"You change value of non-existing property.",@""),
```
lifcio: Not sure if we were using it before, but props for using `NSLocalizedString` :+1:

syncano-ios/SCDataObject+Increment.m
```diff
@@ -0,0 +1,91 @@
...
+                               };
+    return [NSError errorWithDomain:SCDataObjectErrorDomain  code:SCErrorCodeDataObjectNonExistingPropertyName userInfo:userInfo];
+}
+
+- (NSDictionary*)buildParametrsForIncrementQueryForValues:(NSDictionary<NSString*,NSNumber*>*)values withError:(NSError *__autoreleasing *)error {
```
lifcio: typo, `paramet*e*rs`

syncano-ios/SCDataObject+Increment.m
```diff
@@ -0,0 +1,91 @@
...
+        if ([[[self class] propertyKeys] containsObject:propertyName]) {
+            params[propertyName] = @{@"_increment":value};
+        } else {
+            *stop = YES;
+            internalError = [self errorForUnknownProperty:propertyName];
+        }
+    }];
+    if (self.revision) {
+        params[kExpectedRevisionRequestParam] = self.revision;
+    }
```
lifcio: Won't adding it always check revision, even if we didn't provide the block to check it? (self.revision is alway not nil of we downloaded object from Syncano)

syncano-ios/SCDataObject+Increment.m
```diff
@@ -0,0 +1,91 @@
...
+        return nil;
+    }
+    return params;
+}
+
+- (void)fillValues:(NSDictionary<NSString*,NSNumber*>*)values fromResponseObject:(id)responseObject {
+    [values enumerateKeysAndObjectsUsingBlock:^(NSString * _Nonnull propertyName, NSNumber * _Nonnull value, BOOL * _Nonnull stop) {
+        [self setValue:[responseObject valueForKey:propertyName] forKey:propertyName];
+    }];
+    self.updated_at = responseObject[@"updated_at"];
```
lifcio: This will set `NSDate*` variable to `NSString*` value.

lifcio:
I'm not sure about the method naming convention. We have `incrementKey` and `incrementValues`. 

Maybe we should stick to one, and have either `key` and `keys` or `value` and `values`? 

I would opt for `incrementKey` and `incrementKeys`

SyncanoIntegrationTests/DataObjectManipulationsSpec.m
trapi: Well, I'm not sure. beforeAll is for sure more understandable, will change it. These differences are traces of fight with Kiwi and specs for background operations ;)

syncano-ios/SCDataObject+Increment.m
trapi: @jaskooo What do you think? It's copied from updateValue

trapi:
Fixed. Having @jaskooo response, we will be OK to merge.

syncano-ios/SCDataObject+Increment.m
jaskooo: Let's leave it right now.
We will add **forceUpdate** Bool param to save method which will tell us to use this check or no.
The inside normal **saveWithComplition** method we will use **forceUpdate** = YES**

PR #69, opened 2016-02-05
[LIB-490] Fix ability to save files

lifcio:
Instead of blindly refreshing whole object right before saving files
(and by that - removing these files before they were saved) we now
modify response object to contain files that need to be uploaded.

Also made needsToBeUploaded public. I tried using valueForKey and SEL but it felt dirty, and could easily break with potential property name change.

Added FMDB as a dependency - otherwise it doesn’t install in projects using our library.

### DIFF
--- a/syncano-ios/SCDataObject.m
+++ b/syncano-ios/SCDataObject.m
@@ -195,27 +195,12 @@
     }];
 }
 
-- (NSDictionary *)filesForPropertyNamesThatNeedToBeUploaded {
-    NSMutableDictionary *filesForPropertyNamesThatNeedToBeUploaded = [NSMutableDictionary dictionary];
-    NSArray *filesProperties = [[self class] propertiesNamesOfFileClass];
-    for (NSString *filePropertyName in filesProperties) {
-        SCFile * file = (SCFile *)[self valueForKey:filePropertyName];
-        if (file.needsToBeUploaded) {
-            filesForPropertyNamesThatNeedToBeUploaded[filePropertyName] = file;
-        }
-    }
-    return filesForPropertyNamesThatNeedToBeUploaded;
-}
-
-- (id)responseObjectWithFileValuesThatNeedToBeUploaded:(id)responseObject {
-    NSMutableDictionary *responseCopy = [responseObject mutableCopy];
-    [responseCopy setValuesForKeysWithDictionary:[self filesForPropertyNamesThatNeedToBeUploaded]];
-    return responseCopy;
-}
-
 - (void)updateObjectAfterSaveWithDataFromJSONObject:(id)responseObject {
-    id responseObjectWithFileValuesThatNeedToBeUploaded = [self responseObjectWithFileValuesThatNeedToBeUploaded:responseObject];
-    [[SCParseManager sharedSCParseManager] fillObject:self withDataFromJSONObject:responseObjectWithFileValuesThatNeedToBeUploaded];
+    self.objectId = responseObject[@"id"];
+    self.links = responseObject[@"links"];
+    self.revision = responseObject[@"revision"];
+    self.created_at = [[SCConstants SCDataObjectDatesTransformer] transformedValue:responseObject[@"created_at"]];
+    self.updated_at = [[SCConstants SCDataObjectDatesTransformer] transformedValue:responseObject[@"updated_at"]];
 }
 
 - (void)saveFilesUsingAPIClient:(SCAPIClient *)apiClient completion:(SCCompletionBlock)completion {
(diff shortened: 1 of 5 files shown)


lifcio:
@trapi Could you review? Unfortunately with my last fix, I broke something else saving files. Fill with JSON was using response from server with no files or old files and replacing/removing the one we wanted to send, right before calling `saveFilesUsingAPIClient` method.

syncano-ios/SCDataObject.m
```diff
@@ -195,9 +195,32 @@ - (void)saveUsingAPIClient:(SCAPIClient *)apiClient withCompletion:(SCCompletion
...
+
+- (id)responseObjectWithFileValuesThatNeedToBeUploaded:(id)responseObject {
+    NSMutableDictionary *responseCopy = [responseObject mutableCopy];
+    [responseCopy setValuesForKeysWithDictionary:[self filesForPropertyNamesThatNeedToBeUploaded]];
+    return responseCopy;
+}
+
+- (void)updateObjectAfterSaveWithDataFromJSONObject:(id)responseObject {
+    id responseObjectWithFileValuesThatNeedToBeUploaded = [self responseObjectWithFileValuesThatNeedToBeUploaded:responseObject];
+    [[SCParseManager sharedSCParseManager] fillObject:self withDataFromJSONObject:responseObjectWithFileValuesThatNeedToBeUploaded];
```
trapi: I don't like this "fillObject" after save.. It can overwrite user's data. It was introduced before but I missed it.
Consider this scenario:
1. book.title = @"t1"
2. [book save]
3. book.title = @"t2"
4. //Now result form save (point 2) came and book.title is @"t1". Change from 3 is lost

syncano-ios/SCDataObject.m
```diff
@@ -195,9 +195,32 @@ - (void)saveUsingAPIClient:(SCAPIClient *)apiClient withCompletion:(SCCompletion
...
+        SCFile * file = (SCFile *)[self valueForKey:filePropertyName];
+        if (file.needsToBeUploaded) {
+            filesForPropertyNamesThatNeedToBeUploaded[filePropertyName] = file;
+        }
+    }
+    return filesForPropertyNamesThatNeedToBeUploaded;
+}
+
+- (id)responseObjectWithFileValuesThatNeedToBeUploaded:(id)responseObject {
+    NSMutableDictionary *responseCopy = [responseObject mutableCopy];
```
trapi: It's "(id)responseObjectWithFileValuesThatNeedToBeUploaded:(id)responseObject" but you assume quietly here that responseObject is type of NSDictionary*

syncano-ios/SCDataObject.m
```diff
@@ -195,9 +195,32 @@ - (void)saveUsingAPIClient:(SCAPIClient *)apiClient withCompletion:(SCCompletion
     }];
 }
 
+- (NSDictionary *)filesForPropertyNamesThatNeedToBeUploaded {
+    NSMutableDictionary *filesForPropertyNamesThatNeedToBeUploaded = [NSMutableDictionary dictionary];
+    NSArray *filesProperties = [[self class] propertiesNamesOfFileClass];
+    for (NSString *filePropertyName in filesProperties) {
+        SCFile * file = (SCFile *)[self valueForKey:filePropertyName];
+        if (file.needsToBeUploaded) {
```
trapi: Are you 100% sure file is always type of SCFile? I'd rather use [file respondsToSelector]

syncano-ios/SCDataObject.m
lifcio: Yes, but that was intentional on my part - so after saving we always have the newest values from the backend. But I just went through some use cases and I was wrong in how it works in general. Good catch!

syncano-ios/SCDataObject.m
lifcio: Yes, but we actually assume that everywhere :) I can add a check, but if that's a problem we should add them everywhere.

syncano-ios/SCDataObject.m
lifcio: Hmm... `propertiesNamesOfFileClass` gives me properties that should of type `SCFile`, but actually yeah - it's possible someone assigned variable of a different type. Will fix.

syncano-ios/SCParseManager+SCDataObject.m
```diff
@@ -59,7 +59,7 @@ - (void)resolveRelationsToObject:(id)parsedObject withJSONObject:(id)JSONObject
 
 - (void)resolveFilesForObject:(id)parsedObject withJSONObject:(id)JSONObject {
     for (NSString *key in [JSONObject allKeys]) {
-        if ([parsedObject respondsToSelector:@selector(key)]) {
+        if ([parsedObject respondsToSelector:NSSelectorFromString(key)]) {
```
lifcio: That fixed downloading and parsing files, thanks!

PR #71, opened 2016-02-09
[LIB-483] revision mismatch integration tests

jaskooo:


SyncanoIntegrationTests/DataObjectRevisionMismatchSpec.m
```diff
@@ -0,0 +1,89 @@
...
+#import "Syncano.h"
+#import "SCFileManager.h"
+#import "Book.h"
+
+SPEC_BEGIN(DataObjectRevisionMismatch)
+
+describe(@"DataObjectRevisionMismatch", ^{
+    
+    NSDictionary *environment = [[NSProcessInfo processInfo] environment];
+    NSString *apikey = environment[@"API_KEY"];
```
trapi: As @lifcio mentioned somewhere - camelCase

SyncanoIntegrationTests/DataObjectRevisionMismatchSpec.m
```diff
@@ -0,0 +1,89 @@
...
+            __block BOOL _blockFinished;
+            __block BOOL _mismatched;
+            __block NSString *_mismatchDescription;
+
+            //Getting books from API
+            [[Book please] giveMeDataObjectsWithCompletion:^(NSArray *objects, NSError *error) {
+                Book *_book = [objects firstObject];
+                // Savig first of them
+                
+                _book.revision = @(_book.revision.integerValue - 1);
```
trapi: I'm not sure if it's so important but when we have (_book==nil && error!=nil) here then this test fails (so works as expected) but we will get no information where.
Writing such tests (nested blocks) I try to store and check for previous error or at least [[_book should] beNonNil];

PR #75, opened 2016-02-10
[hot-fix] Fetching SCFile

lifcio:
Without this fix, it always returns self.data, which is nil by default (unless `storeDataAfterFetch` was set to `YES`)


lifcio:
Already pushed to master to fix a support issue, but still check please.

jaskooo:
LGTM :) Thanks

PR #76, opened 2016-02-11
[FIX] Swift class names demangling

lifcio:
@jaskooo @trapi My first try on that. I’m not sure if we should do that, or ask users to add `@objc(classname)` before their classes.

Forcing users to do sth before using their class is wrong, not intuitive and easy to miss.

What I suggested is easier for user as doesn’t require any extra step form them but is also and probably at some point might stop working…

Let me know what you think

### DIFF
--- a/syncano-ios/SCParseManager.m
+++ b/syncano-ios/SCParseManager.m
@@ -20,14 +20,50 @@
 
 SINGLETON_IMPL_FOR_CLASS(SCParseManager)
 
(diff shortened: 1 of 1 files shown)


jaskooo:
Users are forced to return the class name for API anyway. If your solution will work for both obj-c and swift it is ok for me. Looks good anyway :)

lifcio:
Yep, works for both ObjC and Swift :)

PR #81, opened 2016-02-12
Hotifx 4.0.10

lifcio:
Fixed using custom profile and custom user class.

Before, `[SCUser currentUser]` was being instantiated before custom classes could have been registered -- because of that `currentUser` was always of class `SCUser` and profile always of class `SCUserProfile` .

@jaskooo @trapi 
I have a problem with tests here, not sure how to write them. 
The issue wasn't discovered in current tests, because we log out user out after each single test (which clears user object saved on disk). 
The issue is only visible when there is a user object on disk. Any suggestions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ Syncano's iOS Library (syncano-ios) is available under the MIT license. See the 
 
 ## Change Log
 --
+* **4.0.10** - 2016-02-12
+    * Fixed using custom class for User and User Profile 
 * **4.0.9** - 2016-02-11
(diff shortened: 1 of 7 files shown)


lifcio:
I'm thinking now - maybe we should also clean the `_currentProfile` variable after each successful call to `registerClassWithUserProfile:`? (not remove it from user defaults, just set the variable to `nil`)

This way, we could theoretically swap user profile classes multiple times during one runtime - doing it currently will crash. Probably not really a common use, but sounds neat :)

jaskooo:
Code looks good.

PR #82, opened 2016-02-13
Create new user without logging him in automatically

lifcio:
Added a new static method in SCUser to allow registering users without login him automatically. It also returns the newly created user so you can, for exemple, update its profile right away. Useful in cases where an admin of a private system wants to register users, because they can't use the app without being invited.

by @patricktremblay1

### DIFF
--- a/syncano-ios/SCConstants.h
+++ b/syncano-ios/SCConstants.h
@@ -10,12 +10,14 @@
 @class SCTrace;
 @class SCWebhookResponseObject;
 @class SCChannelNotificationMessage;
+@class SCUser;
 
 typedef void (^SCAPICompletionBlock)(NSURLSessionDataTask *task, id responseObject, NSError *error);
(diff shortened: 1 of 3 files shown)


lifcio:
@patricktremblay1 I made small changes - name of the blog, and I hid method using `API Client`, to add one using `Syncano` object instead. Let me know if it looks good.

@jaskooo @trapi I need to add test, so you can wait with review.

patricktremblay1:
@lifcio Looks good, thanks.

jaskooo:
I think this method shoulde be named `addNewUserWithUserName:paasword:completionBlock:`

lifcio:
Ok, thanks, I think it makes sense - so we can keep method `register` to auto-login user, and `add` to just add him, without login. 
I will write tests this week and try to merge it before Sunday.

lifcio:
I'll close it for now until tests are done and reopen when they are committed to this branch.

lifcio:
@patricktremblay1 It should be on cocoapods on Monday

PR #83, opened 2016-02-15
Saving empty relations fix

lifcio:
Fixes saving empty relations. Before, when object was equal to `nil`, both
```objc
[relatedObject isKindOfClass:[SCDataObject class]]
```
and
```objc
if (objectId)
```
were returning NO, causing errors and not saving objects properly.


lifcio:
@jaskooo Take a look

jaskooo:
LGTM :)

PR #84, opened 2016-02-16
[LIB-484] add missing predicates

trapi:
It doesn't support SCPredicateIsOperator because it would be useful only for saving Views to local storage. Unfortunately such action breaks relation - only related object id is saved, not the whole object. In such circumstances SCPredicateIsOperator support is useless.


lifcio:
LGTM, I like extensive tests :)

jaskooo:
LGTM

PR #86, opened 2016-02-17
Fix: When user doesn't hold reference to the object or file's object …

trapi:
When user doesn't hold reference to the object or file then returned file's path is nil


jaskooo:
LGTM

lifcio:
LGTM

PR #93, opened 2016-02-18
Fix error on nil predicate

lifcio:


SyncanoIntegrationTests/LocalStorageSpec.m
```diff
@@ -161,6 +173,11 @@
             [[_error should] beNil];
         });
         
+        it(@"should not throw exception for nil predicate", ^{
+            SCPredicate* nilPredicate = nil;
```
jaskooo: Unused variable :)

jaskooo:
Except this one comment looks good :)

SyncanoIntegrationTests/LocalStorageSpec.m
lifcio: Oops, moved it to the block, and didn't remove here, thanks!

PR #97, opened 2016-02-25
Lib 497 register device for push notifications

jaskooo:


syncano-ios/SCDevice.h
```diff
@@ -0,0 +1,44 @@
...
+//  Created by Jan Lipmann on 25/02/16.
+//  Copyright © 2016 Syncano. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "SCConstants.h"
+
+@class Syncano;
+
+@interface SCDevice : NSObject
```
lifcio: Why not use Mantle here? Would be in line with all other classes we have.

syncano-ios/SCDevice.m
```diff
@@ -0,0 +1,101 @@
...
+
+- (NSDictionary *)metadata {
+    if (!_metadata) {
+        _metadata = [NSDictionary new];
+    }
+    return _metadata;
+}
+
+- (void)setMetadataObject:(id)object forKey:(nonnull NSString *)key {
+    NSMutableDictionary *metadata = [[self metadata] mutableCopy];
```
lifcio: Wouldn't it be better for `_metadata` to be mutable? Not sure if making an object copy every time we add a new key is efficient - sounds like a perfect job for mutable dictionary.

PR #101, opened 2016-03-02
[LIB-471] Hotfix for NSDate saving

lifcio:


lifcio:
Will make another one with updated fix.

PR #103, opened 2016-03-02
Merge Hotfixes

lifcio:
Fix saving NSDate objecs
Fix polling a channel blocks making any other requests


syncano-ios/SCDataObject.m
```diff
@@ -54,6 +77,11 @@ + (NSValueTransformer *)JSONTransformerForKey:(NSString *)key {
         [key isEqualToString:@"updated_at"]) {
         return [SCConstants SCDataObjectDatesTransformer];
     }
+    
+    Class aClass = [self classForKey:key];
+    if ([aClass isSubclassOfClass:[NSDate class]]) {
```
lifcio: Probably could have used that for `created_at` and `updated_at` as well. @jaskooo what do you think?

syncano-ios/SCDataObject.m
```diff
@@ -54,6 +77,11 @@ + (NSValueTransformer *)JSONTransformerForKey:(NSString *)key {
         [key isEqualToString:@"updated_at"]) {
```
jaskooo: You mean to add `created_at` here?

lifcio: I meant that in theory 

```objc
Class aClass = [self classForKey:key];
 +    if ([aClass isSubclassOfClass:[NSDate class]]) {
``` 

should also handle `created_at` and `updated_at` as well - but we may as well leave it as it is, as these are special values.

PR #105, opened 2016-03-07
Lib 489 

jaskooo:
As a developer I want to have identical objects in relations

### DIFF
--- a/syncano-ios/SCParseManager+SCDataObject.m
+++ b/syncano-ios/SCParseManager+SCDataObject.m
@@ -20,7 +20,7 @@
     if(parsedobject == nil) {
         return parsedobject;//possible error in parsing
     }
-    
+    [referencesStore addDataObject:parsedobject];
     [self resolveRelationsToObject:parsedobject withJSONObject:JSONObject];
     [self resolveFilesForObject:parsedobject withJSONObject:JSONObject];
     return parsedobject;
@@ -35,11 +35,12 @@
     NSNumber *relatedObjectId = JSONObject[@"value"];
     if (relatedObjectId) {
         //we have only id
-        id relatedObject = [[objectClass alloc] init];
-        [relatedObject setValue:relatedObjectId forKey:@"objectId"];
-        if ([referencesStore containsObject:relatedObject]) {
-            return [referencesStore member:relatedObject];
+        id relatedObject = [referencesStore getObjectById:relatedObjectId];
+        if (relatedObject) {
+            return relatedObject;
         } else {
+            relatedObject = [[objectClass alloc] init];
+            [relatedObject setValue:relatedObjectId forKey:@"objectId"];
             [referencesStore addDataObject:relatedObject];
         }
         return relatedObject;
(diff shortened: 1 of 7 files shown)


syncano-ios/SCParseManager+SCDataObject.m
```diff
@@ -37,6 +37,11 @@ - (id)relatedObjectOfClass:(__unsafe_unretained Class)objectClass fromJSONObject
         //we have only id
         id relatedObject = [[objectClass alloc] init];
         [relatedObject setValue:relatedObjectId forKey:@"objectId"];
+        if ([referencesStore containsObject:relatedObject]) {
+            return [referencesStore member:relatedObject];
```
lifcio: This won't work properly from what I see - you first allocate memory for a new object, set a value on it, and then you return one from referenceStore - but the one in reference store was never updated.
You should create new object only if there isn't one in memory.

syncano-ios/SCParseManager.h
```diff
@@ -9,6 +9,9 @@
 #import "SCMacros.h"
 #import "SCConstants.h"
 #import "Mantle/NSError+MTLModelException.h"
+#import "SCReferencesStore.h"
+
+static SCReferencesStore  * referencesStore;
```
lifcio: I would make it a local variable. If it's being used in Singleton, it would work the same, but would also allow in future to have different ParseManagers (and different referenceStore) per Syncano Instance (which might be needed, cause ids are unique only within 1 instance).

jaskooo: Yes. I had the same thoughts today. I do not know yet how to wire this to syncano instance.

lifcio: For now I would just move it into being a local variable inside `SCParseManager` and tackle different parse managers per instance in a different ticket.

jaskooo: Yes. Good idea

syncano-ios/SCReferencesStore.m
```diff
@@ -0,0 +1,34 @@
...
+- (instancetype)init {
+    self = [super init];
+    if (self) {
+        self.store = [[NSMapTable alloc] initWithKeyOptions:NSMapTableStrongMemory valueOptions:NSMapTableWeakMemory capacity:0];
+    }
+    return self;
+}
+
+- (void)addDataObject:(SCDataObject *)dataObject {
+    [self.store setObject:dataObject forKey:dataObject.objectId];
```
lifcio: Just to be sure, I would also check here for id, and if nil, just return nil. 

In other places in code we often deal with objects without ID's so might be easy to forget about how reference store works -- if we add this check here, we should be good.

lifcio: Sorry, it doesn't return anything -- so just do nothing if it doesn't have an ID, right now it will probably throw an exception.

lifcio:
Other than the one above - looks good!

syncano-ios/SCReferencesStore.m
jaskooo: Right!

lifcio:
LGTM

PR #116, opened 2016-03-16
[LIB]-515

lifcio:
Added Script Endpoint and Script Socket API
Added JSON file for Script Trace
Added private interface for SCAPI client so it can be used in other
files
Deprecated methods in SCTrace that use CodeBox endpoints
SCTrace now by default uses API 1.1
Added/fixed tests for changes made

In general - I think that’s good enough. End goal should be for SCAPI client to accept API version argument, so we don’t have to create new API client for each API version, but I want to tackle that in a separate task.


lifcio:
@jaskooo Can you take a look?

jaskooo:
LGTM :)

PR #124, opened 2016-03-24
Update AFNetworking to version 3.0.0

SoneeJohn:


lifcio:
Thanks for the PR! We'll look into that more after Easter, but in general we have a plan to switch library to use AFNetworking 3.0 everywhere.
It will need a bit more testing, but thanks for a nudge :)

jaskooo:
Hello @SoneeJohn thanks for contribution :) Here we have included your changes: https://github.com/Syncano/syncano-ios/pull/137. Thanks again :)

PR #140, opened 2016-04-20
Hotfix - Add SCPredicate for `whereKeyDoesNotExist`

lifcio:


lifcio:
Waiting with merge until we have updated test config.

PR #141, opened 2016-04-20
Fix - adding auth challenge

lifcio:
Needed for working with staging SSL. Without it, we get -999 Cancelled error


lifcio:
Tested locally, merging.

PR #142, opened 2016-04-25
Lib 523 annotations

jaskooo:


syncano-ios/SCAPIClient+SCFile.h
```diff
@@ -18,7 +20,7 @@
  *
  *  @return NSURLSessionDataTask
  */
-+ (NSURLSessionDataTask *)downloadFileFromURL:(NSURL *)fileURL withCompletion:(SCAPIFileDownloadCompletionBlock)completion;
++ (nullable NSURLSessionDataTask *)downloadFileFromURL:(NSURL *)fileURL withCompletion:(nullable SCAPIFileDownloadCompletionBlock)completion;
```
lifcio: NSURLSessionDataTask will never be null (it's not nullable in Foundation)

syncano-ios/SCAPIClient+SCFile.h
```diff
@@ -30,8 +32,9 @@
  *
  *  @return NSURLSessionDownloadTask
  */
-+ (NSURLSessionDownloadTask *)downloadFileFromURL:(NSURL *)fileURL
++ (nullable NSURLSessionDownloadTask *)downloadFileFromURL:(NSURL *)fileURL
```
lifcio: It's not nullable in NSURLSession

syncano-ios/NSData+MimeType.h
```diff
@@ -9,5 +9,5 @@
 #import <Foundation/Foundation.h>
 
 @interface NSData (MimeType)
-- (NSString *)mimeTypeByGuessing;
+- (nullable NSString *)mimeTypeByGuessing;
```
lifcio: It's not nullable, you always return string (check implementation)

syncano-ios/NSDictionary+JSONString.h
```diff
@@ -9,5 +9,5 @@
 #import <Foundation/Foundation.h>
 
 @interface NSDictionary (JSONString)
--(NSString*) sc_jsonStringWithPrettyPrint:(BOOL) prettyPrint error:(NSError *__autoreleasing*)error;
+-(nullable NSString*) sc_jsonStringWithPrettyPrint:(BOOL) prettyPrint error:( NSError *  __autoreleasing _Nullable * _Nullable)error;
```
lifcio: Return type is not nullable, in case of error we return `@{}`.
error parameter doesn't require `_Nullable` passed twice - this is the default mode anyway:
```
The particular type NSError ** is so often used to return errors via method parameters that it is always assumed to be a nullable pointer to a nullable NSError reference.
```
(https://developer.apple.com/swift/blog/?id=25)

syncano-ios/NSString+JSONDictionary.h
```diff
@@ -9,5 +9,5 @@
 #import <Foundation/Foundation.h>
 
 @interface NSString (JSONDictionary)
-- (NSDictionary *) sc_jsonDictionary:(NSError *__autoreleasing*)error;
+- (nullable NSDictionary *) sc_jsonDictionary:(NSError *  __autoreleasing _Nullable * _Nullable)error;
```
lifcio: Return type ok, not needed for NSError (as above)

syncano-ios/NSString+MD5.h
```diff
@@ -9,5 +9,5 @@
 #import <Foundation/Foundation.h>
 
 @interface NSString (MD5)
-- (NSString *)sc_MD5String;
+- (nullable NSString *)sc_MD5String;
```
lifcio: Not nullable, we never return `nil`.

syncano-ios/SCAPIClient.h
```diff
@@ -23,7 +25,7 @@
  *
  *  @return SCAPIClient object
  */
-+ (SCAPIClient *)apiClientForSyncano:(Syncano *)syncano;
++ (nullable SCAPIClient *)apiClientForSyncano:(Syncano *)syncano;
```
lifcio: I think it shouldn't be nullable, it's `nil` only when NSObject alloc method fails (and NSObject `init` does not return optional value)

syncano-ios/SCAPIClient_SCAPIClient.h
```diff
@@ -27,10 +29,11 @@
  *
  *  @return SCAPIClient
  */
-- (instancetype)initWithApiVersion:(SCAPIVersion)apiVersion apiKey:(NSString *)apiKey instanceName:(NSString *)instanceName;
+- (nullable instancetype)initWithApiVersion:(SCAPIVersion)apiVersion apiKey:(NSString *)apiKey instanceName:(NSString *)instanceName;
```
lifcio: This is null only if NSObject `init` fails, I don't think it should be nullable in this case

syncano-ios/SCAPIClient_SCAPIClient.h
```diff
@@ -27,10 +29,11 @@
...
 
 /**
  @deprecated This method is deprecated. Please use initWithApiVersion:apiKey:instanceName: instead when initializing API Client to use it with Syncano
  */
-- (instancetype)initWithBaseURL:(NSURL *)url apiKey:(NSString *)apiKey instanceName:(NSString *)instanceName DEPRECATED_MSG_ATTRIBUTE("Use initWithApiVersion:apiKey:instanceName: method instead.");
+- (nullable instancetype)initWithBaseURL:(NSURL *)url apiKey:(NSString *)apiKey instanceName:(NSString *)instanceName DEPRECATED_MSG_ATTRIBUTE("Use initWithApiVersion:apiKey:instanceName: method instead.");
```
lifcio: As above

syncano-ios/SCChannel.h
```diff
@@ -41,7 +43,7 @@
  *
  *  @return SCChannel instance
  */
-- (instancetype)initWithName:(NSString *)channelName;
+- (nullable instancetype)initWithName:(NSString *)channelName;
```
lifcio: As before - check e.g. AFNetworking library, its `init`s are not nullable.

syncano-ios/SCChannel.h
```diff
@@ -51,7 +53,7 @@
  *
  *  @return SCChannel instance
  */
-- (instancetype)initWithName:(NSString *)channelName andDelegate:(id<SCChannelDelegate>)delegate;
+- (nullable instancetype)initWithName:(NSString *)channelName andDelegate:(nullable id<SCChannelDelegate>)delegate;
```
lifcio: As above

syncano-ios/SCChannel.h
```diff
@@ -61,7 +63,7 @@
  *
  *  @return SCChannel instance
  */
-- (instancetype)initWithName:(NSString *)channelName lastId:(NSNumber *)lastId andDelegate:(id<SCChannelDelegate>)delegate;
+- (nullable instancetype)initWithName:(NSString *)channelName lastId:(NSNumber *)lastId andDelegate:(nullable id<SCChannelDelegate>)delegate;
```
lifcio: As above

syncano-ios/SCChannel.h
```diff
@@ -72,7 +74,7 @@
  *
  *  @return SCChannel instance
  */
-- (instancetype)initWithName:(NSString *)channelName room:(NSString *)room andDelegate:(id<SCChannelDelegate>)delegate;
+- (nullable instancetype)initWithName:(NSString *)channelName room:(NSString *)room andDelegate:(nullable id<SCChannelDelegate>)delegate;
```
lifcio: As above

syncano-ios/SCChannel.h
```diff
@@ -84,7 +86,7 @@
  *
  *  @return SCChannel instance
  */
-- (instancetype)initWithName:(NSString *)channelName lastId:(NSNumber *)lastId room:(NSString *)room andDelegate:(id<SCChannelDelegate>)delegate;
+- (nullable instancetype)initWithName:(NSString *)channelName lastId:(NSNumber *)lastId room:(NSString *)room andDelegate:(nullable id<SCChannelDelegate>)delegate;
```
lifcio: As above

syncano-ios/SCChannel.h
```diff
@@ -104,7 +106,7 @@
  *  @param payload    NSDictionary payload
  *  @param completion completion block
  */
-- (void)publishToChannelWithPayload:(NSDictionary *)payload completion:(SCChannelPublishCompletionBlock)completion;
+- (void)publishToChannelWithPayload:(nullable NSDictionary *)payload completion:(nullable SCChannelPublishCompletionBlock)completion;
```
lifcio: I think we can require a payload here, otherwise it doesn't make sense to to even use this method. I would remove `nullable` from `payload`.

syncano-ios/SCChannel.h
```diff
@@ -113,20 +115,21 @@
  *  @param payload    NSDictionary payload
  *  @param completion completion block
  */
-- (void)publishToChannelInSyncano:(Syncano *)syncano withPayload:(NSDictionary *)payload completion:(SCChannelPublishCompletionBlock)completion;
+- (void)publishToChannelInSyncano:(Syncano *)syncano withPayload:(nullable NSDictionary *)payload completion:(nullable SCChannelPublishCompletionBlock)completion;
```
lifcio: As above

syncano-ios/SCChannel.h
```diff
@@ -113,20 +115,21 @@
...
 
 /**
  *  Gets channel history using singleton Syncano instance
  *
  *  @param completion completion blog
  */
-- (void)getChannelHistoryWithCompletion:(SCChannelHistoryCompletionBlock)completion;
+- (void)getChannelHistoryWithCompletion:(nullable SCChannelHistoryCompletionBlock)completion;
```
lifcio: Only purpose of this method is to get the history. If someone wants to pass `nil` it doesn't make any sense. I would make this block required and remove `nullable`

syncano-ios/SCChannel.h
```diff
@@ -113,20 +115,21 @@
...
 
 /**
  *  Gets channel history using provided Syncano instance
  *
  *  @param syncano    Syncano instance
  *  @param completion completion block
  */
-- (void)getChannelHistoryFromSyncano:(Syncano *)syncano completion:(SCChannelHistoryCompletionBlock)completion;
+- (void)getChannelHistoryFromSyncano:(Syncano *)syncano completion:(nullable SCChannelHistoryCompletionBlock)completion;
```
lifcio: As above

syncano-ios/SCChannelDelegate.h
```diff
@@ -11,6 +11,6 @@
 
 @protocol SCChannelDelegate <NSObject>
 
-- (void)chanellDidReceivedNotificationMessage:(SCChannelNotificationMessage *)notificationMessage;
+- (void)chanellDidReceivedNotificationMessage:(nonnull SCChannelNotificationMessage *)notificationMessage;
```
lifcio: Doesn't look from the code as this can be ever `nil`

syncano-ios/SCChannelHistoryResponse.h
```diff
@@ -10,10 +10,10 @@
...
 @interface SCChannelHistoryResponse : NSObject
-@property (nonatomic,copy) NSString *next;
-@property (nonatomic,copy) NSString *prev;
-@property (nonatomic,copy) NSArray<SCChannelNotificationMessage*> *objects;
+@property (nullable,nonatomic,copy) NSString *next;
+@property (nullable,nonatomic,copy) NSString *prev;
+@property (nullable,nonatomic,copy) NSArray<SCChannelNotificationMessage*> *objects;
 
-- (instancetype)initWithJSONObject:(id)JSONObject;
+- (nullable instancetype)initWithJSONObject:(nonnull id)JSONObject;
```
lifcio: Init

syncano-ios/SCChannelNotificationMessage.h
```diff
@@ -11,12 +11,12 @@
...
+@property (nullable,nonatomic,copy) NSDate *createdAt;
+@property (nullable,nonatomic,copy) NSDictionary *author;
 @property (nonatomic) SCChannelNotificationMessageAction action;
-@property (nonatomic,copy) NSDictionary *payload;
-@property (nonatomic,copy) NSDictionary *metadata;
+@property (nullable,nonatomic,copy) NSDictionary *payload;
+@property (nullable,nonatomic,copy) NSDictionary *metadata;
 
-- (instancetype)initWithJSONObject:(id)JSONObject;
+- (nullable instancetype)initWithJSONObject:(nonnull id)JSONObject;
```
lifcio: Init

syncano-ios/SCCodeBox.h
```diff
@@ -30,7 +32,7 @@
  *  @param params     params
  *  @param completion completion block
  */
-+ (void)runCodeBoxWithId:(NSNumber *)codeBoxId params:(NSDictionary *)params completion:(SCCodeBoxCompletionBlock)completion;
++ (void)runCodeBoxWithId:(NSNumber *)codeBoxId params:(NSDictionary *)params completion:(nullable SCCodeBoxCompletionBlock)completion;
```
lifcio: Params are not required (all data may be in source code or config), make it nullable

syncano-ios/SCCodeBox.h
```diff
@@ -40,6 +42,7 @@
  *  @param syncano    provided Syncano instance
  *  @param completion completion block
  */
-+ (void)runCodeBoxWithId:(NSNumber *)codeBoxId params:(NSDictionary *)params onSyncano:(Syncano *)syncano completion:(SCCodeBoxCompletionBlock)completion;
++ (void)runCodeBoxWithId:(NSNumber *)codeBoxId params:(NSDictionary *)params onSyncano:(Syncano *)syncano completion:(nullable SCCodeBoxCompletionBlock)completion;
```
lifcio: Params should be nullable

syncano-ios/SCCompoundPredicate.h
```diff
@@ -23,7 +25,7 @@
  *
  *  @return SCCompoundPredicate
  */
-+ (instancetype)compoundPredicateWithPredicates:(NSArray *)predicates;
++ (nullable instancetype)compoundPredicateWithPredicates:(NSArray *)predicates;
```
lifcio: Init, not nullable

syncano-ios/SCCompoundPredicate.h
```diff
@@ -32,7 +34,7 @@
  *
  *  @return SCCompoundPredicate
  */
-- (instancetype)initWithPredicates:(NSArray *)predicates;
+- (nullable instancetype)initWithPredicates:(NSArray *)predicates;
```
lifcio: Init, not nullable

syncano-ios/SCConstants.h
```diff
@@ -123,13 +123,16 @@ typedef NS_ENUM(NSUInteger, SCAPIVersion) {
...
+NS_ASSUME_NONNULL_BEGIN
+
 @interface SCConstants : NSObject
 + (SCDataObjectPermissionType)dataObjectPermissiontypeByString:(NSString *)typeString;
 + (SCChannelPermisionType)channelPermissionTypeByString:(NSString *)typeString;
 + (SCChannelType)channelTypeByString:(NSString *)typeString;
-+ (NSString *)socialAuthenticationBackendToString:(SCSocialAuthenticationBackend)backend;
-+ (NSValueTransformer *)SCDataObjectPermissionsValueTransformer;
-+ (NSValueTransformer *)SCDataObjectDatesTransformer;
++ (nullable NSString *)socialAuthenticationBackendToString:(SCSocialAuthenticationBackend)backend;
```
lifcio: There's no default value that would return `nil` - it's not nullable

syncano-ios/SCConstants.h
```diff
@@ -123,13 +123,16 @@ typedef NS_ENUM(NSUInteger, SCAPIVersion) {
...
++ (nullable NSValueTransformer *)SCDataObjectPermissionsValueTransformer;
++ (nullable NSValueTransformer *)SCDataObjectDatesTransformer;
 + (SCChannelNotificationMessageAction)channelNotificationMessageActionByString:(NSString *)actionString;
-+ (NSURL *)baseURLForAPIVersion:(SCAPIVersion)apiVersion;
-@end
\ No newline at end of file
++ (nullable NSURL *)baseURLForAPIVersion:(SCAPIVersion)apiVersion;
```
lifcio: It's never `nil`

syncano-ios/SCDataObject+LocalStorage.h
```diff
@@ -16,23 +16,23 @@
...
+- (void)saveToLocalStorageWithCompletion:(nullable SCCompletionBlock)completion;
 
 #pragma mark - Fetch -
 /**
  *  Fetches all data objects from local database
  *
  *  @param completionBlock completion block
  */
-+ (void)fetchAllObjectsFromLocalStorageWithCompletionBlock:(SCDataObjectsCompletionBlock)completionBlock;
++ (void)fetchAllObjectsFromLocalStorageWithCompletionBlock:(nullable SCDataObjectsCompletionBlock)completionBlock;
```
lifcio: This method doesn't make sense if we pass `nil`

syncano-ios/SCDataObject+LocalStorage.h
```diff
@@ -41,12 +41,12 @@
  *
  *  @param completion completion block
  */
-- (void)generateInsertQueryWithCompletion:(SCLocalStorageGenerateQueryStringCompletionBlock)completion;
+- (void)generateInsertQueryWithCompletion:(nullable SCLocalStorageGenerateQueryStringCompletionBlock)completion;
```
lifcio: Doesn't make much sense with `nil`

syncano-ios/SCDataObject+LocalStorage.h
```diff
@@ -41,12 +41,12 @@
...
 
 /**
  *  Generates DELETE query for current SCDataObject
  *
  *  @param completion completion block
  */
-- (void)generateDeleteQueryWithCompletion:(SCLocalStorageGenerateQueryStringCompletionBlock)completion;
+- (void)generateDeleteQueryWithCompletion:(nullable SCLocalStorageGenerateQueryStringCompletionBlock)completion;
```
lifcio: Doesn't make much sense with `nil`

syncano-ios/SCDataObject+Properties.h
```diff
@@ -15,12 +15,12 @@
  *
  *  @return NSArray of strings
  */
-+ (NSArray *)propertiesNamesOfFileClass;
++ (nonnull NSArray *)propertiesNamesOfFileClass;
```
lifcio: Never `nil` (worst case it's an empty array)

syncano-ios/SCDataObject+Properties.h
```diff
@@ -15,12 +15,12 @@
...
 
 /**
  *  Returns classes of all properties inside a SCDataObject
  *
  *  @return NSDictionary with property name as a key and property class name as a value
  */
-+ (NSDictionary *)classesOfProperties;
++ (nonnull NSDictionary *)classesOfProperties;
```
lifcio: Never nil (worst case it's an empty dictionary)

syncano-ios/SCDataObject.h
```diff
@@ -48,21 +50,21 @@
...
-+ (NSDictionary *)extendedPropertiesMapping;
++ (nullable NSDictionary *)extendedPropertiesMapping;
 
 /**
  *  Returns SCPlease instance for singleton Syncano
  *
  *  @return SCPlease instance
  */
-+ (SCPlease *)please;
++ (nullable SCPlease *)please;
```
lifcio: Can this be nil?

syncano-ios/SCDataObject.h
```diff
@@ -71,7 +73,7 @@
  *
  *  @return SCPlease instance
  */
-+ (SCPlease *)pleaseFromSyncano:(Syncano *)syncano;
++ (nullable SCPlease *)pleaseFromSyncano:(Syncano *)syncano;
```
lifcio: As above

lifcio:
Wow, that was huge... I didn't even check 1/3 of changes. 

1. In general, as a good rule, try to use audited regions in each file, this way any nullable will have to be explicit (`NS_ASSUME_NONNULL_BEGIN`), even if file has only one or two functions.
2. In `init`, unless you or superclass explicitly return `nil` value, I would assume it's not nullable. For good practice, check even `NSString` or `NSData` (or AFNetworking file) headers.

Hope I can finish the review before next sprint :)

syncano-ios/NSDictionary+JSONString.h
jaskooo: I've noticed that apple uses` NSError **` as nonnull like in `NSData` : `+ (nullable instancetype)dataWithContentsOfURL:(NSURL *)url options:(NSDataReadingOptions)readOptionsMask error:(NSError **)errorPtr;` I think we should follow this convention

dancio:
`22` days PR :> what's happening? close it or merge it

lifcio:
In SCPlease.h (cannot comment in there, because code wasn't change on the methods I want to comment):
` - (void)giveMeDataObjectsWithPredicate:(id<SCPredicateProtocol>)predicate parameters:(NSDictionary *)parameters completion:(SCDataObjectsCompletionBlock)completion;` should accept nil for parameters -- otherwise we need another method where we can pass just predicate, without parameters. 
Same for enumerate -- both predicate and parameters should be nullable, otherwise there is not way to provide one or the other (not both) -- and actually I might want to just enumerate just all elements, without filtering them.

lifcio:
Other than that looks good, I run it against few of my apps, and it seems these changes only helped, so after two small changes described above, good to merge!

PR #145, opened 2016-05-04
[LIB-644] Sweep user data (logout) before login

jaskooo:


syncano-ios/SCUser.m
```diff
@@ -153,7 +155,11 @@ + (void)loginWithSocialBackend:(SCSocialAuthenticationBackend)backend authToken:
 }
 
 - (void)logout {
-    [[self class] removeUserFromDefaults];
+    [[self class] sweepLoggedUserData];
+}
+
++ (void)sweepLoggedUserData {
```
lifcio: Maybe `sweepLoggedInUserData` ? So `loggedIn`, not just `logged`

lifcio:
other than the comment above, lgtm

PR #147, opened 2016-05-17
Lib 546 template tests

jaskooo:


syncano-ios/SCPleaseForTemplate.m
```diff
@@ -0,0 +1,58 @@
...
+
++ (SCPleaseForTemplate *)pleaseInstanceForDataObjectWithClass:(Class)dataObjectClass forTemplate:(NSString *)templateName forSyncano:(Syncano *)syncano {
+    SCPleaseForTemplate* instance = (SCPleaseForTemplate*)[self pleaseInstanceForDataObjectWithClass:dataObjectClass forSyncano:syncano];
+    instance.templateName = templateName;
+    return instance;
+}
+
+- (void)giveMeDataWithParameters:(NSDictionary*)parameters completion:(SCCustomResponseCompletionBlock)completion {
+    SCAPIClient *apiClient = [[self apiClient] copy];
+    apiClient.responseSerializer = [AFHTTPResponseSerializer serializer];
```
lifcio: Does this always make response object be of type `NSData`? If yes, maybe we could change our `SCCustomResponseCompletionBlock` to always have `NSData` inside? (instead of vague `id`)

SyncanoIntegrationTests/TemplatesIntegrationTestsSpec.m
```diff
@@ -0,0 +1,49 @@
...
+    [Author registerClass];
+    
+    context(@"Fetch data with template", ^{
+        
+        __block NSError *_error;
+        __block BOOL _blockFinished = NO;
+        __block id _responseObject;
+        it(@"should text", ^{
+            
+            [[Book pleaseForTemplate:@"itworks"] giveMeDataWithParameters:nil completion:^(id responseObject, NSError *error) {
```
lifcio: Like commented in the other place - if this was `NSData` it wouldn't be require to cast it manually.

jaskooo: `SCCustomResponseCompletionBlock` is used in several places we cannot directly assume that it will be `NSData`.

lifcio: Ok, right, I see it would be problematic with custom script endpoints - sometimes they return NSData, sometimes NSDictionary or something else.

But in this case it would be always NSData, as you always declare standard serializer:
```swift
apiClient.responseSerializer = [AFHTTPResponseSerializer serializer];
```
and there is no way to change it. Maybe add a new callback, specifically for templates? If it's always NSData, we shouldn't make people guess it.

On the other hand, `id` works fine if we added option to pass custom `responseSerializer`, so then it could be json, xml, html etc.

Both work fine - always NSData, or `id` + custom serializer. What we have now is not the right solution in my opinion.

PR #148, opened 2016-05-17
[LIB-705] Fix spelling of a method name in SCChannelDelegate

jaskooo:


syncano-ios/SCChannelDelegate.h
```diff
@@ -10,7 +10,8 @@
 #import "SCChannelNotificationMessage.h"
 
 @protocol SCChannelDelegate <NSObject>
-
+- (void)chanellDidReceiveNotificationMessage:(SCChannelNotificationMessage *)notificationMessage;
```
lifcio: chaNNeL - double `n` and single `l`

PR #154, opened 2016-06-01
Lib 648 geopoint

jaskooo:


lifcio:
Looks good. We should add new file to Syncano.h as well, so it's added automatically for everyone.

PR #155, opened 2016-06-06
Lib 651 geopoint filtering

jaskooo:


syncano-ios/SCPredicate.h
```diff
@@ -358,4 +358,14 @@ extern NSString *const SCPredicateIsOperator;
...
 
 @end
+
+@class SCGeoPoint;
+
+extern NSString *const SCPredicateNearOpeartor;
+
+@interface SCPredicate (GeoPoint)
++ (SCPredicate *)whereKey:(NSString *)key nearGeoPoint:(SCGeoPoint *)geopoint;
++ (SCPredicate *)whereKey:(NSString *)key nearGeoPoint:(SCGeoPoint *)geopoint withinMiles:(double)maxDistance;
```
lifcio: I would keep the name consistent with other predicates - isEqual, isGreaterThan etc.
So in here it would be `isWithinMiles`

syncano-ios/SCPredicate.h
```diff
@@ -358,4 +358,14 @@ extern NSString *const SCPredicateIsOperator;
...
++ (SCPredicate *)whereKey:(NSString *)key nearGeoPoint:(SCGeoPoint *)geopoint withinKilometers:(double)maxDistance;
```
lifcio: `isWithinKilometers`

syncano-ios/SCPredicate.h
jaskooo: i think it should rather be `isNearGeoPoint:withinMiles:`

syncano-ios/SCPredicate.h
jaskooo: maybe `whereKey:isNearGeoPoint:withinKilometers:` ?

syncano-ios/SCPredicate.h
lifcio: Agree :+1:

syncano-ios/SCPredicate.h
lifcio: Sounds good :+1:

PR #157, opened 2016-06-08
Lib 749 geopoint tests

jaskooo:


lifcio:
Looks good. Did anything change with other tests? Array/object etc - removed and new lines look the same.

jaskooo:
I just put them into separate contexts. That's all

PR #158, opened 2016-06-10
Lib 707 relations

jaskooo:
Here is a class that uses relation
```
#import "Syncano.h"

@interface BookStore : SCDataObject
@property (nonatomic,retain) SCRelation *books;
@end
```
You can use it like this

```
    BookStore *bookstore = [BookStore new];
    bookstore.books = [SCRelation relationWithTargetClassName:@"book"];
    [[Book please] giveMeDataObjectsWithParameters:@{SCPleaseParameterPageSize : @3} completion:^(NSArray * _Nullable books, NSError * _Nullable error) {
        [bookstore.books setMembers:books];
        [bookstore saveWithCompletionBlock:^(NSError * _Nullable error) {
            
        }];
    }];
```

### DIFF
--- a/syncano-ios/SCRelation.m
+++ b/syncano-ios/SCRelation.m
@@ -20,8 +20,21 @@
 
 @implementation SCRelation
 
++ (SCRelation *)relationWithTargetClassName:(NSString *)targetClassName {
+    SCRelation *relation = [[[self class] alloc] initWithTargetClassName:targetClassName];
+    return relation;
+}
+
+- (instancetype)initWithTargetClassName:(NSString *)targetClassName {
+    self = [super init];
+    if (self) {
+        self.targetClassName = targetClassName;
+    }
+    return self;
+}
+
 -(instancetype)initWithDictionary:(NSDictionary *)dictionaryValue error:(NSError *__autoreleasing *)error targetClassName:(NSString *)targetClassName {
-    self = [super initWithDictionary:dictionaryValue error:error];
+    self = [super init];
     if (self) {
         self.membersIds = [[dictionaryValue[@"value"] sc_arrayOrNil] mutableCopy];
         self.targetClassName = targetClassName;
@@ -37,8 +50,12 @@
     return self;
 }
 
++ (NSDictionary *)JSONKeyPathsByPropertyKey {
+    return [NSDictionary mtl_identityPropertyMapWithModel:[self class]];
+}
+
 - (BOOL)isKindOfTargetClass:(SCDataObject *)member {
-    return [NSStringFromClass([member class]) isEqualToString:self.targetClassName];
+    return [[NSStringFromClass([member class]) lowercaseString] isEqualToString:self.targetClassName];
 }
 
 - (SCPlease *)please {
--- a/syncano-ios/SCRelation.m
+++ b/syncano-ios/SCRelation.m
@@ -1,0 +1,52 @@
+//
+//  SCRelation.m
+//  syncano-ios
+//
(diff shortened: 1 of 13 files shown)


jaskooo:
`[[BookStore please] giveMeDataObjectsWithCompletion:^(NSArray * _Nullable objects, NSError * _Nullable error) {
        BookStore *bookstore = (BookStore *)[objects firstObject];
        if(bookstore.books == nil) {
             bookstore.books = [SCRelation relationWithTargetClassName:@"book"];
        }
        [[Book please] giveMeDataObjectsWithParameters:@{SCPleaseParameterPageSize : @3} completion:^(NSArray * _Nullable books, NSError * _Nullable error) {
            for (Book *book in books) {
                [bookstore.books addDataObject:book];
            }
            [bookstore saveWithCompletionBlock:^(NSError * _Nullable error) {
                
            }];
        }];
        
    }];`

syncano-ios/SCRelation.m
```diff
@@ -0,0 +1,52 @@
...
+    if (![self isKindOfTargetClass:object]) {
+        return;
+    }
+    NSNumber *objectId = object.objectId;
+    if (![self.membersIds containsObject:objectId]) {
+        return;
+    }
+    [self.membersIds removeObject:objectId];
+}
+@end
```
jaskooo: @lifcio I understand that those two methods should be changed to add and remove objects directly to API with completion block and there should be added method to set localy objects to releation like `books.members = @[book1, book2, book3]` ?

lifcio: Yeah, adding/removing should go to API directly. Locally should work normally, as normal array, integer, string etc - it should just replace whole relations array.

jaskooo:
@lifcio done.

lifcio:
SCRelations class looks good. I was thinking about using normal NSArray that could hold just SCDataObjects, but I tried it a bit and it gets confusing - how to know which NSArray send as array, which as relations, you need to inspect etc... So SCRelations class looks good.

Can you add here a few examples of how to use SCPlease for relations with and without predicates? It wasn't straightforward to me.

jaskooo:
I think we should reduce relation's please class to only fetch objects like this:
```
__block BookStore *bookstore;
    [[BookStore please] giveMeDataObjectsWithCompletion:^(NSArray * _Nullable objects, NSError * _Nullable error) {
        bookstore = [objects firstObject];
        [[bookstore.books please] giveMeDataObjectsWithCompletion:^(NSArray * _Nullable objects, NSError * _Nullable error) {
            
        }];
    }];
```

jaskooo:
This please is created to make queries on Data Objects on `target` class and the only thing added is SCPredicate added to reduce results to ids from relation

jaskooo:
@lifcio as for predicates. Here all the magic is happening:
`SCPleaseForRelation.m`

```
- (void)setPredicate:(id<SCPredicateProtocol>)predicate {
    SCPredicate *membersPredicate = [SCPredicate whereKey:@"id" inArray:_relationMembers];
    SCCompoundPredicate *compoundPredicate;
    if (predicate) {
        if ([predicate isKindOfClass:[SCCompoundPredicate class]]) {
            compoundPredicate = (SCCompoundPredicate *)predicate;
            [compoundPredicate addPredicate:membersPredicate];
        } else {
            compoundPredicate = [SCCompoundPredicate compoundPredicateWithPredicates:@[predicate,membersPredicate]];
        }
        [super setPredicate:compoundPredicate];
    } else {
        [super setPredicate:membersPredicate];
    }
}
```

PR #163, opened 2016-06-16
Lib 781 add filter to paging

jaskooo:


lifcio:
LGTM.

I am a little bit confused by having both `parameters` and `queryParameters`. They are basically the same, just one is processed and one is not. And sometimes you save one, but set other to nil, only to later save it again...

Maybe `resolveQueryParameters` method should be synchronous? and then we could get rid of `self.queryParameters` and always pass `[self resolveQueryParameters:self.parameters]`?

jaskooo:
@lifcio That was a first thought, but i wanted to reduce count of resolving query function runs. But if we agree that resolving on the fly will be better it is no problem for me to rollback :)

PR #164, opened 2016-06-22
Lib 654 cache key

jaskooo:


syncano-ios/NSDictionary+CacheKey.m
```diff
@@ -0,0 +1,26 @@
...
+//  Created by Jan Lipmann on 22/06/16.
+//  Copyright © 2016 Syncano. All rights reserved.
+//
+
+#import "NSDictionary+CacheKey.h"
+#import "SCConstants.h"
+
+@implementation NSDictionary (CacheKey)
++ (NSDictionary *)dictionaryWithCacheKey:(NSString *)cacheKey {
+    return @{SCPleaseParameterCacheKey : cacheKey};
```
lifcio: Check for nil here, otherwise it will throw an exception.

syncano-ios/NSDictionary+CacheKey.m
```diff
@@ -0,0 +1,26 @@
...
+}
+
+- (NSDictionary *)dictionaryByAddingCacheKey:(NSString *)cacheKey {
+    if (self) {
+        NSMutableDictionary *mutable = [self mutableCopy];
+        mutable[SCPleaseParameterCacheKey] = cacheKey;
+        return [NSDictionary dictionaryWithDictionary:mutable];
```
lifcio: Only matter of preference, but you could just use `[mutable copy]` which also return immutable verison. A bit shorter ;)

syncano-ios/NSString+PathManipulations.m
```diff
@@ -0,0 +1,19 @@
...
+
+#import "NSString+PathManipulations.h"
+
+@implementation NSString (PathManipulations)
+- (NSString *)pathStringByAppendingQueryString:(NSString *)queryString {
+    if (![queryString length]) {
+        return self;
+    }
+    return [NSString stringWithFormat:@"%@%@%@", self,
+            [self rangeOfString:@"?"].length > 0 ? @"&" : @"?", queryString];
```
lifcio: To make it more readable I would rather compare location to `NSNotFound` instead of length to 0

syncano-ios/SCAPIClient.m
```diff
@@ -329,3 +330,24 @@ - (BOOL)reachable {
 }
 
 @end
+
+@implementation SCAPIClient (CacheKey)
+
+- (void)checkAndResolveCacheKeyExistanceInPayload:(NSDictionary *)payload forPath:(NSString *)path completion:(void(^)(NSString *path, NSDictionary *payload))completion {
+    if ([[payload allKeys] containsObject:SCPleaseParameterCacheKey]) {
```
lifcio: No need to get array of all keys, just use `objectForKey` and check if the returned value is `nil` or not

syncano-ios/SCAPIClient.m
```diff
@@ -329,3 +330,24 @@ - (BOOL)reachable {
...
+        NSString *cacheKey = payload[SCPleaseParameterCacheKey];
+        NSMutableDictionary *mutablePayload = [payload mutableCopy];
+        [mutablePayload removeObjectForKey:SCPleaseParameterCacheKey];
+        NSString *cacheKeyQueryString = [NSString stringWithFormat:@"%@=%@",SCPleaseParameterCacheKey,cacheKey];
+        NSString *pathWithCacheKey = [path pathStringByAppendingQueryString:cacheKeyQueryString];
+        if (completion) {
```
lifcio: Maybe instead of checking twice (here and again later), check for completion at the beginning and if it's nil just return. No need to make a link if there's no callback to be called.

syncano-ios/SCPleaseForView.h
```diff
@@ -44,4 +44,10 @@ NS_ASSUME_NONNULL_BEGIN
 + (SCPlease *)pleaseInstanceForDataObjectWithClass:(Class)dataObjectClass forView:(NSString*)viewName forSyncano:(Syncano *)syncano;
 
 @end
+
+@interface SCPleaseForView (Cache)
+- (void)giveMeDataObjectsfromCacheWithKey:(NSString *)cacheKey withCompletion:(SCDataObjectsCompletionBlock)completion;
```
lifcio: from -> From

PR #169, opened 2016-06-29
[LIB-790] Addd autoregister of all SCDataObject classes while creatin…

jaskooo:
…g syncano object

### DIFF
--- a/syncano-ios/SCRegisterManager.m
+++ b/syncano-ios/SCRegisterManager.m
@@ -9,6 +9,8 @@
 #import "SCRegisterManager.h"
 #import "SCDataObject+Properties.h"
 #import "SCParseManager.h"
+#import <objc/runtime.h>
+#import "SCUserProfile.h"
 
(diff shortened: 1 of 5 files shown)


syncano-ios/SCRegisterManager.m
```diff
@@ -22,6 +24,46 @@ + (NSMutableArray *)registeredClasses;
...
+    
+    free(classes);
+    
+    return result;
+}
+
++ (void)autoregisterDataObjectSubclasses {
+    NSArray *classes = ClassGetSubclasses([SCDataObject class]);
+    for (Class classToRegister in classes) {
+        if ([classToRegister isSubclassOfClass:[SCUserProfile class]]) {
```
lifcio: Why don't we register SCUserProfile subclass? It should be registered as well, in case someone has a reference from e.g. "Book" to "user_profile".
And we could actually use it to also register it as a profile subclass, instead of just doing nothing.

PR #171, opened 2016-07-02
Lib 797 api version update for Data Endpoint

jaskooo:


lifcio:
LGTM

syncano-ios/SCPleaseForDataEndpoint.m
```diff
@@ -43,7 +45,9 @@ + (SCPlease *)pleaseInstanceForDataObjectWithClass:(Class)dataObjectClass fordat
 }
 
 - (void)getDataObjectFromAPIWithParams:(NSDictionary*)queryParameters completion:(SCDataObjectsCompletionBlock)completion {
-    [[self apiClient] getDataObjectsFromdataEndpointWithName:self.dataEndpointName params:queryParameters completion:^(NSURLSessionDataTask * _Nonnull task, id  _Nullable responseObject, NSError * _Nullable error) {
+    SCAPIClient *apiClient = [self.apiClient copy];
+    apiClient.apiVersion = SCAPIVersion_1_1;
```
lifcio: Actually, now I see this won't work. Base URL isn't changed after setting API Version, as base URL is read-only. So we need to make new API Client here e.g.
```objc
SCAPIClient *apiClient_v1_1 = [[SCAPIClient alloc] initWithApiVersion:SCAPIVersion_1_1 apiKey:apiClient.apiKey instanceName:apiClient.instanceName];
```

jaskooo: Hmm weird it should work, `apiVersion` is not readonly as i know ```@property (nonatomic, assign) SCAPIVersion apiVersion;
```

jaskooo: I will change it like you suggest to be consistant with `SCScript`

lifcio: apiVersion is not readonly, but baseURL is... So one you set api version in init, base url is already set in stone and cannot be changed.
(or maybe could be changed, but we would need to override something from AFNetworking - didn't really look into that)

PR #172, opened 2016-07-04
Lib 789 save object befor add relation

jaskooo:


lifcio:
Looks good. Though I would extract somewhere this code:
```objc
[self saveMeIfNeededWithAPIClient:apiClient completion:^(NSError * _Nullable error) {
        typeof(self) __weak selfWeak = self;
        [apiClient PATCHWithPath:[self path] params:params completion:^(NSURLSessionDataTask *task, id responseObject, NSError *error) {
            if(error) {
                if(completion) {
                    completion(error);
                }
                if(revisionMismatchBlock) {
                    [error checkIfMismatchOccuredWithCompletion:revisionMismatchBlock];
                }
                return;
            }
            
            [selfWeak fillKeys:keys fromResponseObject:responseObject];
            
            if(completion) {
                completion(nil);
            }
            if(revisionMismatchBlock) {
                revisionMismatchBlock(NO,nil);
            }
        }];
    }];
```
cause it's the same in all 3 methods - not reason to just copy-case, instead of creating a new method out of it.

PR #175, opened 2016-07-11
Fix saving files with custom Syncano object

lifcio:
Before the fix, it always tried to used shared instance - which not
always exists. Method received APIClient and should just pass it
forward.


jaskooo:
LGTM :)

PR #179, opened 2016-07-14
Lib 807 file upload don't change object's permissions

jaskooo:


syncano-ios/SCAPIClient.m
```diff
@@ -204,7 +204,7 @@ - (void)runRequest:(SCRequest *)request {
         SCUploadRequest *uploadRequest = (SCUploadRequest *)request;
         NSString *propertyName = uploadRequest.propertyName;
         NSData *fileData = uploadRequest.fileData;
-        [self postUploadTaskWithPath:path propertyName:propertyName fileData:fileData completion:requestFinishedBlock];
+        [self patchUploadTaskWithPath:path propertyName:propertyName fileData:fileData completion:requestFinishedBlock];
```
lifcio: I would add here checking for the method in the request -- right now it's ignored, so setting anything in SCUploadRequest `init` doesn't make any sense:
```
self = [super initWithPath:path method:SCRequestMethodPATCH params:nil callback:callback save:save];
```
Since we can handle now both POST and REQUEST, I would use either one based on `uploadRequest.method`, and if it's different than POST and PATCH - throw an error.

jaskooo: As you can see below upload task is treated individually that's why when you change method while initializing it always took `POST`. Now it will always take `PATCH`:
```
if ([request isKindOfClass:[SCUploadRequest class]]) {
        SCUploadRequest *uploadRequest = (SCUploadRequest *)request;
        NSString *propertyName = uploadRequest.propertyName;
        NSData *fileData = uploadRequest.fileData;
        [self patchUploadTaskWithPath:path propertyName:propertyName fileData:fileData completion:requestFinishedBlock];
    } else {
        switch (method) {
            case SCRequestMethodGET:
                [self getTaskWithPath:path params:params completion:requestFinishedBlock];
                break;
            case SCRequestMethodPOST:
                [self postTaskWithPath:path params:params completion:requestFinishedBlock];
                break;
            case SCRequestMethodPUT:
                [self putTaskWithPath:path params:params completion:requestFinishedBlock];
                break;
            case SCRequestMethodPATCH:
                [self patchTaskWithPath:path params:params completion:requestFinishedBlock];
                break;
            case SCRequestMethodDELETE:
                [self deleteTaskWithPath:path params:params completion:requestFinishedBlock];
                break;
            default:
                break;
        }
    }
```

jaskooo: It is not ignored in other requests

lifcio: Exactly, that's why it doesn't make sense to pass method in SCUploadRequest if we ignore it.

jaskooo: Ok then. I will change the initializer for it ;)

lifcio: If it would be Swift we could switch on both class and method ;) Here it would be more complex unfortunately. 

I would rather add a second switch inside `if ([request isKindOfClass:[SCUploadRequest class]]) {` or even better - make two method out of it, each with it's own switch. One for running upload requests, one for running other requests.

This way we will be able to change method used for running requests easily, without manipulating the SCAPIClient.

lifcio:
LGTM :)

PR #181, opened 2016-07-15
Lib 498 send push

jaskooo:


syncano-ios/SCDevice.h
```diff
@@ -14,11 +14,16 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface SCDevice : MTLModel
+@interface SCDevice : MTLModel <MTLJSONSerializing>
 @property (nullable,nonatomic,readonly) NSString * deviceToken;
```
lifcio: I would mark it as deprecated. No need to hold a variable that doesn't even exist on the backend. And remove it in the next release.

syncano-ios/SCPleaseForPush.h
```diff
@@ -0,0 +1,32 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+
+@interface SCPleaseForPush : NSObject
+@property (nonatomic,retain) Syncano *syncano;
+@property (nonatomic,retain) NSArray<SCDevice*>* devices;
+@property (nonatomic) SCPushPleaseEnvironment environment;
+
++ (SCPleaseForPush *)pleaseForSyncano:(Syncano *)syncano environment:(SCPushPleaseEnvironment)environment devices:(NSArray<SCDevice*>*)devices;
```
lifcio: No please using default Syncano shared instance? Shouldn't we also have an interface on SCDevice for sending push messages? E.g.
```objc
[[SCDevice pleaseForEnvironment:...] sendMessage:@"aaa"]; //by default only current device
[[SCDevice pleaseForEnvironment:... devices:@[list]] sendMessage:@"aaa"]; //or pass list of devices
```
Otherwise it's weird, different than the rest of API we have. Nowhere else we access `please` directly, it's always attached to another object.

jaskooo: SCDevice will have its own pleas for fetch devices from server.

lifcio: But it will be `SCPleaseForPush`? So maybe we're mixing things here. If we use `please` only for fetching, maybe we shouldn't send messages using it? And add some other class, e.g. `SCPushMessage` for that purpose?

PR #186, opened 2016-07-25
[LIB-665] Added methods to pass SCUserProfile while registering new user

jaskooo:


lifcio:
Late - but shouldn't 
```
+ (void)registerWithUsername:(NSString *)username password:(NSString *)password usingAPIClient:(SCAPIClient *)apiClient completion:(SCCompletionBlock)completion {
```
now be using the new method
```
+ (void)registerWithUsername:(NSString *)username password:(NSString *)password userProfile:(SCUserProfile *)userProfile usingAPIClient:(SCAPIClient *)apiClient completion:(SCCompletionBlock)completion {
```
? Of course this would require a change in the second one, to allow passing nil values for user profile.

Or is that on purpose that we have 2 methods and one of them is non null for user profile? I'm worried about duplicate code in those 2 methods.

jaskooo:
@lifcio you are right. Done

PR #188, opened 2016-07-27
[LIB-665] removed duplicated private register method

jaskooo:


lifcio:
Looks good. You also need to add `nullable` to user profile in the declaration.

jaskooo:
There's no declaration in header for this method. It is private method for internal use only

lifcio:
I meant 
```objc
+ (void)registerWithUsername:(NSString *)username password:(NSString *)password userProfile:(SCUserProfile *)userProfile completion:(nullable SCCompletionBlock)completion;
```
and 
```objc
+ (void)registerWithUsername:(NSString *)username password:(NSString *)password userProfile:(SCUserProfile *)userProfile inSyncano:(Syncano *)syncano completion:(nullable SCCompletionBlock)completion;
```

jaskooo:
ok, but in those methods that you pointed `userProfile` is not nullable. It should be provided

lifcio:
Ahh, ok, right. I missed that! LGTM then.

PR #189, opened 2016-08-11
Lib 812 batch

jaskooo:


syncano-ios/SCBatch.m
```diff
@@ -0,0 +1,113 @@
...
+
+#import "SCBatch.h"
+#import "SCBatchRequest.h"
+#import "SCParseManager+SCDataObject.h"
+#import "SCAPIClient.h"
+#import "Syncano.h"
+#import "NSError+SCBatch.h"
+#import "SCBatchResponseItem.h"
+
+static NSInteger maxRequestsCount = 50;
```
lifcio: I think it would be better if it stayed in SCConstants.h and be public this way.

jaskooo: This is a private var, there is no point to move it to public, because API describes max requests as 50

syncano-ios/NSError+SCBatch.m
```diff
@@ -0,0 +1,20 @@
...
+//
+//  Created by Jan Lipmann on 03/08/16.
+//  Copyright © 2016 Syncano. All rights reserved.
+//
+
+#import "NSError+SCBatch.h"
+#import "SCConstants.h"
+
+@implementation NSError (SCBatch)
++ (NSError*)maxRequestExceededErrorForMaxRequestNumber:(NSInteger)max {
```
lifcio: It doesn't hurt to have it as a variable, but if we know it's a const, and we don't use a value from server in here, it could just take no parameters (unless we plan to change it later to actually parse server response).

syncano-ios/SCBatch.h
```diff
@@ -0,0 +1,25 @@
...
+
+#import <Foundation/Foundation.h>
+#import "SCConstants.h"
+
+@class Syncano,SCDataObject;
+
+@interface SCBatch : NSObject
+
++ (SCBatch *)batch;
++ (SCBatch *)batchForSyncano:(Syncano *)syncano;
```
lifcio: :+1: Outside of the scope of this ticket, but we have different naming in different objects - sometimes its `forSyncano`, sometimes `fromSyncano` - we should fix that later.

syncano-ios/SCBatch.m
```diff
@@ -0,0 +1,113 @@
...
+
+- (void)sendWithCompletion:(SCBatchRequestCompletionBlock)completion {
+    [self.apiClient POSTWithPath:@"batch/" params:[self encodedRequests] completion:^(NSURLSessionDataTask * _Nonnull task, id  _Nullable responseObject, NSError * _Nullable error) {
+        if(completion) {
+            completion([self handelResponse:responseObject],error);
+        }
+    }];
+}
+
+- (NSArray<SCBatchResponseItem *>*)handelResponse:(id)responseObject {
```
lifcio: `handle`, not `handel`

syncano-ios/SCBatch.h
lifcio: What I mean is - if we want later to make everything to be `forSyncano` then this is ok. If we want to stick to `fromSyncano` we should change this method.

syncano-ios/SCBatch.m
```diff
@@ -0,0 +1,113 @@
...
+    NSMutableArray<SCBatchResponseItem *> *decodedItems = [NSMutableArray new];
+    
+    NSInteger index = 0;
+    
+    for (NSDictionary *item in responseObject) {
+        SCBatchResponseItem *decodedItem = [SCBatchResponseItem itemWithJSONDictionary:item classToParse:[self responseObjectClassForRequestWithIndex:index]];
+        [decodedItems addObject:decodedItem];
+        index ++;
```
lifcio: Why a space between variable and `++`? ;)

syncano-ios/SCBatchRequest.m
```diff
@@ -0,0 +1,42 @@
...
+
++ (SCBatchRequest *)requestWithMethod:(SCRequestMethod)method path:(NSString *)path payload:(NSDictionary *)payload {
+    SCBatchRequest *request = [SCBatchRequest new];
+    request.method = method;
+    request.path = path;
+    request.payload = payload;
+    return request;
+}
+
++ (SCBatchRequest *)requestWithMethod:(SCRequestMethod)method path:(NSString *)path payload:(NSDictionary *)payload responseObjectClass:(Class)responseObjectClass {
```
lifcio: Shouldn't this method create a new batch request object object, and `+ (SCBatchRequest *)requestWithMethod:(SCRequestMethod)method path:(NSString *)path payload:(NSDictionary *)payload` use it by passing `nil` as `request.responseObjectClass` object?

syncano-ios/SCBatch.m
lifcio: Yes, and by making it public we would allow users to check what is max number of requests, by directly checking the variable. Otherwise they will know only if they will exceed it and get an error.

syncano-ios/SCBatch.h
jaskooo: Here we should use `ForSyncano` because of context

jaskooo:
Done

lifcio:
Ok, LGTM

PR #190, opened 2016-08-29
[LIB-829] retry after implementation

jaskooo:


syncano-ios/SCAPIClient.m
```diff
@@ -173,20 +173,48 @@ - (void)runRequest:(SCRequest *)request {
...
+                NSNumber *retryAfter = nil;
+                if ([task.response isKindOfClass:[NSHTTPURLResponse class]]) {
+                    NSHTTPURLResponse *response =  (NSHTTPURLResponse *)task.response;
+                    if ([response respondsToSelector:@selector(allHeaderFields)]) {
+                        NSDictionary *allHeaderFields = response.allHeaderFields;
+                        retryAfter = allHeaderFields[@"retry_after"];
+                    }
+                }
+                if (retryAfter != nil) {
+                    [self.requestQueue enqueueRequest:request onTop:YES];
```
jaskooo: I do not know what we should do when user will turn off the app while we will be waiting for delay to finish ? @lifcio any thoughts?

